PR #1, opened 2021-08-23
Update package name to @activeviam/activeui-migration

antoinetissier:


Nouzbe (commented):
The change from `migration` to `activeui-migration` is good 👍 

I think we will have to remove the `@activeviam/` scope though, otherwise everyone at ActiveViam and its customers will have to toggle the registry for this scope back and forth between our Artifactory and NPM whenever installing this dependency, which will be annoying.

antoinetissier:
Reassigning to you after our discussion:
- we are not sure whether the bundle already includes all dependencies, or if running `yarn install activeui-migration` would also try to fetch its dependencies, in which case an artifactory login would be needed.
- even if the bundle does include all dependencies and no credentials are required, our target users are paying customers.

=> It makes more sense to keep everything in artifactory, and therefore to keep the `@activeviam` scope as well.

antoinetissier:
And reassigning back to me to implement your suggestion @Nouzbe, because in fact we do want everyone - not just paying customers - to be able to use those scripts: it will be required for atoti (community) users wanting to migrate content.

PR #10, opened 2021-12-13
UI-6728 - Migrate katex formulas

rgosse:
Adds handling of katex formula migration:

Old notation
```
$ inline formula $
$$ block formula $$
```

becomes
```
`katex inline formula`

```katex
block formula
 \```
```


Nouzbe (approved):
Looks good!

PR #12, opened 2021-12-20
UI-6692 - Sanitize Mdx before deriving mapping from it

Nouzbe:
ActiveUI 4 has a small bug when the user collapses and then re-expands a member: in this case, it generates a suboptimal Mdx. This comes with a small symptom in the UI: in order to re-collapse it from there, the user needs to click twice. This is a minor bug and it would be expensive to fix in ActiveUI 4. I do not believe that we will ever do it.

But this suboptimal Mdx unfortunately also makes the content migration CLI throw an error and fail to migrate the impacted widget. Specifically, [_getAxisHierarchyFields.ts](https://github.com/activeviam/activeui/blob/ee193ae46817a6bd6efaacfbb89638643066be88/packages/widget/src/utils/_getAxisHierarchyFields.ts) (called by `deriveGenericMappingFromMdx`, called by `deriveMappingFromMdx`) throws errors. See [UI-6692](https://support.activeviam.com/jira/browse/UI-6692).

This PR fixes that by sanitizing the Mdx in order to remove the useless (and dangerous) parts in it before `deriveMappingFromMdx` is called, in the case of table widgets but also charts and KPI widgets.

As a reviewer, testing that this pull request works is not trivial, I recommend trusting the numerous added tests (and the already existing ones).

Note: the first commit is a migration to ActiveUI 5.0.6 latest. It could be moved to a separate PR.


Nouzbe:
@messaooudi could you give me a review here?

src/_findErroneousUnionNode.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Returns whether `mdx` represents an erroneous union node.
+ */
+export function _findErroneousUnionNode(
+  mdx: Mdx,
+  { cube, levelsOnAxis }: { cube: Cube; levelsOnAxis: LevelCoordinates[] }
+): { path: (string | number)[]; match: ErroneousUnionNode } | undefined {
+  const result = findDescendant(mdx, (node) => {
+    if (!isMdxFunction(node, "union") || node.arguments.length < 2) {
+      // The erroneous node is an union function, with at least 2 arguments.
```
messaooudi, line 32: Reading this comment inside the if statements feels misleading, the if statement represent the negation of the comment ! I would negate the sentence or move it before the if statement with a `should form`
`// The erroneous node is not a union function, with at least 2 arguments.`
or
`// The erroneous node should be an union function, with at least 2 arguments.`

src/_findErroneousUnionNode.ts
```diff
@@ -0,0 +1,78 @@
...
+    ) {
+      // All its arguments are crossjoins.
+      return false;
+    }
+
+    let numberOfCrossjoinsYieldingAllCombinationOfMembers = 0;
+    let numberOfCrossjoinsRepresentingAnExpandedMember = 0;
+
+    // This cast is safe. See early return above.
+    (node.arguments as MdxFunction[]).forEach((crossjoin) => {
```
messaooudi, line 47: Using a `type guard function` in the above if statement to assert that arguments are functions would spare this cast.

src/_findErroneousUnionNode.ts
```diff
@@ -0,0 +1,78 @@
...
+  const result = findDescendant(mdx, (node) => {
+    if (!isMdxFunction(node, "union") || node.arguments.length < 2) {
+      // The erroneous node is an union function, with at least 2 arguments.
+      return false;
+    }
+
+    if (
+      !node.arguments.every((argument) => isMdxFunction(argument, "crossjoin"))
+    ) {
+      // All its arguments are crossjoins.
```
messaooudi, line 39: same as above.

src/_fixErroneousExpansionMdx.test.ts
```diff
@@ -0,0 +1,484 @@
...
+              )
+            )
+          )
+        ) ON ROWS
+        FROM [EquityDerivativesCube]
+        CELL PROPERTIES VALUE, FORMATTED_VALUE, BACK_COLOR, FORE_COLOR, FONT_FLAGS"
+    `);
+  });
+
+  it("removes the problematic parts of the Mdx of a widget where a user collapsed and then re-expanded 2 different members in ActiveUI 4 (collapsing all of them, then re-expanding all of them)", () => {
```
Nouzbe, line 112: While testing the new function, I discovered that collapsing and re-expanding actually contains several bugs in ActiveUI 4.

If you do it on several members, then:
- collapsing then re-expanding each one before moving on to the next yields a bad mdx (see test case above)
- collapsing several in a row, then re-expanding them all yields an even worst mdx (this test case)

src/_fixErroneousExpansionMdx.ts
```diff
@@ -0,0 +1,62 @@
...
+    return mdx;
+  }
+
+  return produce<MdxSelect>(mdx, (draft) => {
+    draft.axes.forEach((axis) => {
+      const levelsOnAxis = getLevels(axis, { cube });
+
+      let erroneousUnionNode:
+        | { path: (string | number)[]; match: ErroneousUnionNode }
+        | undefined = undefined;
```
messaooudi, line 39:
```suggestion
let erroneousUnionNode: ReturnType<typeof _findErroneousUnionNode>
```

src/_fixErroneousExpansionMdx.test.ts
```diff
@@ -0,0 +1,484 @@
...
+        NON EMPTY Hierarchize(
+          Crossjoin(
+            Hierarchize(
+              Union(
+                Hierarchize(
+                  DrilldownLevel(
+                    [Currency].[Currency].[ALL].[AllMember]
+                  )
+                ),
+                [Currency].[Currency].[ALL].[AllMember].[EUR]
```
Nouzbe, line 169: This sanitized Mdx is still bad: ideally, this node should be removed too.

That being said, while being suboptimal, this Mdx does work with ActiveUI 5. See https://github.com/activeviam/activeui/pull/2056. So I do not think that we should make `activeui-migration` more complex in order to cater for it.

src/_fixErroneousExpansionMdx.ts
Nouzbe: I saw you started introducing this syntax in ActiveUI and hesitated making a comment. Let's talk about it here.

I am not in favor of seeing this syntax grow. I have 2 arguments in particular:

1. I do not believe that `ReturnType` is well handled by [react-api-documenter](https://github.com/activeviam/react-api-documenter), so accepting to use it either means:
  a. changing react-api-documenter to make it support it
  b. introducing a convention that using it for internal parts of ActiveUI is fine, but using it for exported functions (or types transitively used by exported functinos) is not.
  c. live with the fact that parts of our documentation will become less useful.

In my previous, I went with b, but I am not satisfied with it: if it grows, I know that I will spend time mentioning it (in average multiple times) to each individual author, and introduce avoidable friction in reviews.

2. Is it really more helpful to the reader? Like, is it not better to have the function actually say what it requires, instead of telling you _"go and look to the definition of this other function in order to know what I take in"_. Think about a future where many types are defined this way. It means that you will generally not only have to go look at a separate function's signature, but actually even have to do it several times in a row, if functionA depends on the return type of functionB, which depends on the return type of ...

src/_findErroneousUnionNode.ts
Nouzbe: The first proposal does not work, as it conveys the opposite of reality here. The second could work, but I don't like it as much.
The reality here is that: _"erroneous node = node to update = union function whose arguments are all crossjoins."_. Maybe I could rewrite the comment that way.

About the _"should form"_, I always avoid it. My reason is that it creates confusion by nature (it being conditional), just like _"might"_. For instance, if you are debugging some code with a comment that say:
```
// `a` is a number.
```
and you realize that during runtime, `a` is in fact not a number, then you know that there **is** a problem with `a`.
Whereas if the comment had said:
```
// `a` should be a number.
```
then you will think that there **might be** a problem with `a`.

And I want you to know that there is a problem, not that there might be one.

In short, I believe that the most direct / unambiguous way of writing is always better, and that the more confused way is always worst.

src/_findErroneousUnionNode.ts
Nouzbe: There already is a typeguard used here (`isMdxFunction`). But as it is used on an array, using `arr.every(typeguard)`, TypeScript cannot tell that `node.arguments` is an array of the type ensured by `isMdxFunction`.

I might be able to make it work without the case if I replace the `.forEach` by a for loop.

src/_fixErroneousExpansionMdx.ts
messaooudi: I agree with the last part as a general guideline , however I can think of cases where a given type definition will be repeated several times across the codebase , in that case I would argue that such a redundancy will become difficult to handle.

In this particular case , I think of `erroneousUnionNode` as a simple object that need no type definition of its own, since it's the result of what ever the function returns ( the need for explicitly typing the object comes from the style of the while loop chosen later on ).

Nouzbe:
Thank you for the review @messaooudi 👍

src/_findErroneousUnionNode.ts
messaooudi: Can't something like this make it work : 
```tsx
const isUnionNodeWithCrossJoinArguments = (
  node: Mdx
): node is MdxFunction & {
  arguments: [
    MdxFunction & { name: "crossjoin" },
    MdxFunction & { name: "crossjoin" }
  ];
}
```

src/_fixErroneousExpansionMdx.ts
Nouzbe: I think that there is a middleground which satisfies us both:
- avoid ReturnType
- avoid redefining the same type multiple times.

This can be achieved by simply putting a name on every type. I will go for it here.

src/_findErroneousUnionNode.ts
Nouzbe: I could do it this way. And actually, this is exactly what I used to call `ErroneousUnionNode`. But I prefer the name you give it: it is more nature-oriented (less role-oriented).

src/_findErroneousUnionNode.ts
```diff
@@ -10,16 +10,35 @@ import { _doesCrossjoinRepresentAnExpandedMember } from "./_doesCrossjoinReprese
...
   name: "Union";
   arguments: (MdxFunction & {
     name: "Crossjoin";
   })[];
 }
 
+/**
+ * A node matched during an Mdx traversal process, and the path leading to it within the traversed mdx.
+ */
+export type MatchAndPath<T extends Mdx> = {
```
Nouzbe, line 25: This type would be useful in ActiveUI as well: for all places using [findDescendant](https://github.com/activeviam/activeui/blob/master/packages/mdx/src/findDescendant.ts).

src/_fixErroneousExpansionMdx.ts
messaooudi: works for me.

messaooudi (approved):
Looks good.

Nouzbe:
@antoinetissier I would appreciate a look from you too, as this change is not trivial.

antoinetissier (commented):
Looks great overall, just some minor suggestions.

src/_doesCrossjoinYieldAllCombinationsOfMembers.ts
```diff
@@ -0,0 +1,29 @@
...
+  // For example:
+  // [Geography].[City].[AllMember].[Paris] // does not count
+  // [Geography].[City].Members // does count
+  return levelsOnAxis.every(({ dimensionName, hierarchyName, levelName }) =>
+    levelsInCrossjoin.some(
+      (levelInCrossjoin) =>
+        levelInCrossjoin.dimensionName === dimensionName &&
+        levelInCrossjoin.hierarchyName === hierarchyName &&
+        levelInCrossjoin.levelName === levelName &&
+        !isMdxCompoundIdentifier(levelInCrossjoin.match)
```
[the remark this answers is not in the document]
Nouzbe: This syntax seems to work actually, in order to retrieve all combination of the members of different levels. See screenshot below. Taking care of it is not complicated. I will make the change and add a test.

image

src/_findErroneousUnionNode.ts
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  if (result === undefined) {
+    return undefined;
+  }
+
+  return {
+    path: result.path,
+    // The implementation of _findErroneousUnionNode ensures that this cast is safe.
+    match: result.match as unknown as UnionNodeWithCrossjoinsAsArguments,
```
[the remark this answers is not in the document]
Nouzbe: Changing the callback passed as argument to `findDescendant` will not avoid the cast, as `findDescendant` returns a generic `Mdx` type anyway.

We might be able to improve this in ActiveUI by typing `findDescendant` as follows:

```
function findDescendant<T extends Mdx = Mdx>(matchFn: (node: Mdx, path: (string | number)[]) => node is T): {
  path: (string | number)[];
  match: T
} | undefined
```

I can still make the comment more specific here.

src/_fixErroneousExpansionMdx.ts
```diff
@@ -0,0 +1,63 @@
...
+        // Remove all expanded members.
+        match.arguments = match.arguments.filter(
+          (crossjoin) =>
+            !_doesCrossjoinRepresentAnExpandedMember(crossjoin, cube)
+        );
+
+        if (match.arguments.length === 1) {
+          // The union has only one argument left. Clean it up.
+          _set(axis, path, match.arguments[0]);
+        }
```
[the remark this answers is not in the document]
Nouzbe: I considered it, but I'd rather not: it seems too advanced to export it.

I am not looking forward to explaining Mdx cleanup to someone outside of the ActiveUI team.

src/_doesCrossjoinYieldAllCombinationsOfMembers.ts
```diff
@@ -23,7 +24,11 @@ export function _doesCrossjoinYieldAllCombinationsOfMembers(
         levelInCrossjoin.dimensionName === dimensionName &&
         levelInCrossjoin.hierarchyName === hierarchyName &&
         levelInCrossjoin.levelName === levelName &&
-        !isMdxCompoundIdentifier(levelInCrossjoin.match)
+        !(
+          isMdxCompoundIdentifier(levelInCrossjoin.match) &&
+          getSpecificCompoundIdentifier(levelInCrossjoin.match, { cube })
+            ?.type === "member"
+        )
```
antoinetissier, line 31: I would find it both more readable and safeproof like this:
```suggestion
        (!isMdxCompoundIdentifier(levelInCrossjoin.match) || getSpecificCompoundIdentifier(levelInCrossjoin.match, { cube })?.type === "level)
```

antoinetissier (approved):

src/_doesCrossjoinYieldAllCombinationsOfMembers.ts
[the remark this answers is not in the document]
antoinetissier: > This syntax seems to work actually
> I don't believe that AUI4 generates this kind of MDX

Yes that's what I meant^^ It's valid so it's nice to support it, although I wouldn't expect to see it much in real life, and it would be custom so there is an "excuse" if it does not work properly when it it migrated.

antoinetissier:
Looks good 👍 Just please consider the last comment before merging.

PR #13, opened 2021-12-21
UI-6727 - Ensure that migrated chart mappings have a key for each attribute, even those who are not mapped to any fields

Nouzbe:
This should fix [UI-6727](https://support.activeviam.com/jira/browse/UI-6727) and [UI-6711](https://support.activeviam.com/jira/browse/UI-6711).

The PR refactors `migrateChart`, in order to cut separate functions for the parts that need a switch in order to do different things for each chart types. It makes the common parts more obvious (such as returning a state which contains query, filters, mapping, ... for instance).

The actual functional change is visible in the test snapshot update: attributes that are mapped to no fields are expressed with an existing key mapped to an empty array, instead of being omitted from the mapping (which does not work in ActiveUI 5).


src/migrateChart.ts
```diff
@@ -10,20 +35,172 @@ import { _getTargetCubeFromServerUrl } from "./_getTargetCubeFromServerUrl";
...
+  legacyChartType: string
+): SerializedDataVisualizationWidgetMapping {
+  const migratedWidgetKey = _getMigratedWidgetKey(legacyChartType);
+
+  if (migratedWidgetKey === undefined) {
+    return legacyMapping;
+  }
+
+  const widgetPlugin = chartPlugins[migratedWidgetKey];
+  const emptyMapping: SerializedDataVisualizationWidgetMapping = {};
```
Nouzbe, line 106: This is the actual fix.

mtadams007 (approved):
Looks good to me.

PR #15, opened 2021-12-23
UI-6890 - Fix migration script failing on text editor widgets when run with nodeJS < v15

Nouzbe:
The used regular expressions already used `g`, making `.replaceAll` directly convertible into `.replace`.

I checked that `migrateTextEditor.test.ts` still passes. It does. Too bad we do not have a CI here yet, which tells it unambiguously to the reviewer.


messaooudi (approved):
Looks good to me.

PR #16, opened 2021-12-29
UI-6905 - Prepare 5.0.7 release

Nouzbe:
This PR bumps the ActiveUI dependency to 5.0.6.

And it bumps the version of `activeui-migration` to 5.0.7.


antoinetissier (commented):
Looks good but there is no need to do a new release per say right ? Or were there changes in the SDK that are needed for the migration script to work properly ?

antoinetissier (approved):

Nouzbe:
> Looks good but there is no need to do a new release per say right ? Or were there changes in the SDK that are needed for the migration script to work properly ?

I am not sure I understand the link between the two questions, but:

- the accelerators team is awaiting the improvements of `activeui-migration`, in particular in order to have a smooth migration of the FRTB accelerator, which should take place at the begining of January. So we should release a new version of it.
- `activeui-migration` should use stable versions of ActiveUI, otherwise people outside ActiveViam will not be able to install it (because the Artifactory imports will not be resolved).

PR #19, opened 2022-04-05
UI-7142 - Use factorized content server initialization utils

antoinetissier:
Depending on `@activeviam/content-server-initialization` to make sure the migration scripts stay in sync with the latest additions in ActiveUI.


Nouzbe (approved):
Looks good 👍

PR #20, opened 2022-04-05
UI-7142 - Bump @activeviam dependencies to 5.0.10

antoinetissier:
It's healthy to be in sync with the latest, and it will also enable to use the factorized utils for initilizing the content server, rather than redefining them in `activeui-migration`.


Nouzbe (commented):
It would be nice to wait for 5.0.10 to be released and do the double bump in one go.

Nouzbe (approved):
Looks good!

PR #21, opened 2022-04-19
UI-7229 - Option to remove widgets of a given type in the CLI

antoinetissier:
The rationale is explained in [UI-7229](https://support.activeviam.com/jira/browse/UI-7229), which comes from a client request at CIBC: [NAS-3463](https://support.activeviam.com/jira/browse/NAS-3463
)

You can review commit by commit.

### DIFF
--- a/src/__snapshots__/migrateUIFolder.test.ts.snap
+++ b/src/__snapshots__/migrateUIFolder.test.ts.snap
@@ -268,6 +268,22 @@ Object {
                 "timestamp": 1607879735685,
               },
             },
+            "eef": Object {
+              "entry": Object {
+                "canRead": true,
+                "canWrite": true,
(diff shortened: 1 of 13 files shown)


src/__test_resources__/legacyUIFolder.ts
```diff
@@ -488,6 +488,32 @@ export const legacyUIFolder = {
                 canWrite: true,
               },
             },
+            "0xb": {
```
antoinetissier, line 491: Fun fact, if the id is not wrapped with quotes, then at runtime it is interpreted as an hexadecimal number, and the id becomes `11` 💩

src/__snapshots__/migrateUIFolder.test.ts.snap
```diff
@@ -2128,6 +2249,18 @@ Object {
         ],
       },
     },
+    "organization_permissions": Object {
```
antoinetissier, line 2252: It seems that I forgot to update the tests last PR.
This has happened for a lot of devs already, and the only way to make it poka-yoke would be to add a CI.
We could even test the CLI in the CI once https://github.com/activeviam/activeui-migration/pull/18 and https://github.com/activeviam/activeui-migration/pull/4 are merged.

src/migrateDashboard.test.ts
```diff
@@ -237,4 +253,39 @@ describe("migrateDashboard", () => {
...
+
+    const dashboard = migrateDashboard(
+      legacyDashboard,
+      servers,
+      keysOfWidgetPluginsToRemove
+    );
+
+    const { content, layout } = dashboard.pages["p-0"];
+    const widgetPluginKeys = _map(content, ({ widgetKey }) => widgetKey);
+    expect(
```
Nouzbe, line 268: It would be nice to add another assertion here with the legacy dashboard state:

```
expect(
  _intersection(widgetKeysInLegacyDashboard, keysOfWidgetPluginsToRemove)
).toHaveLength(1);
```

This way, we are guaranteed that this test will remain useful, even if future devs change the content of mocks.

src/migrateUIFolder.test.ts
```diff
@@ -1,16 +1,73 @@
...
+    const savedWidgets = migratedUIFolder.children!.widgets;
+    const hasRecord = (
+      contentRecord: ContentRecord,
+      recordId: string
+    ): boolean =>
+      _some(
+        contentRecord.children,
+        (child, childId) => childId === recordId || hasRecord(child, recordId)
+      );
+    expect(hasRecord(savedWidgets, "0xb")).toBe(false);
```
Nouzbe, line 40: It would also be nice here to check that the legacy UI folder has a widget with this id.

src/migrateDashboard.ts
```diff
@@ -64,14 +107,42 @@ export function migrateDashboard(
...
     name: legacyDashboardState.name,
     pages,
     pagesOrder: _range(body.pages.length).map((i) => `p-${i}`),
     filters: Object.values(body.filters || {}).flat(),
     queryContext: _migrateContextValues(body.contextValues),
   };
+
+  const deserializedDashboard = deserializeDashboardState(dashboard);
+
+  const dashboardWithWidgetsRemoved = _reduce(
```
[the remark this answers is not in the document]
antoinetissier: That would be ideal, but this is not the signature of `removeWidget`:
```
removeWidget({
    dashboardState,
    pageKey,
    leafKey,
    layoutPath,
});
```
And since `removeWidget` can modify the layout of a given page, then when calling several `removeWidget` in a row, one has to get the latest `dashboardState` at each iteration (rather than the initial `dashboardState` before any removal).
So typically I fell this is a good use case for reduce: you want to feed the intermediate result to the next iteration step.

Or do you have another idea in mind ?

antoinetissier:
> Also, have you been able to try it for real and make sure that the migrated dashboards work as expected?

I have tested that the CLI works well, and that indeed in the output JSON:
- the saved widgets matching the given keys have been removed
- widgets matching the given keys within saved dashboards have been removed

>  I am thinking in particular about edge cases where a page contains a single widget, which is removed. Or even worse: a dashboard contains a single page, which contains a single widget, which is removed.

While I agree that it would be nice to do end-to-end tests for edge cases, here I feel this is almost out of scope: the removal job is the responsability of `removeWidget` from `activeui-sdk`.
If there is a problem with it for some edge cases, then it should be fixed in the SDK, and there should be nothing more to do here than to bump the dependency version.

src/migrateDashboard.ts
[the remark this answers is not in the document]
Nouzbe: I am pushing an attempt. Let me know what you think.

Nouzbe (commented):
I cannot build the repo. I get 40 instances of the following error:

> ERROR in ./node_modules/katex/dist/fonts/KaTeX_Typewriter-Regular.woff 1:4
Module parse failed: Unexpected character '' (1:4)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
(Source code omitted for this binary file)
 @ ./node_modules/css-loader/dist/cjs.js!./node_modules/katex/dist/katex.css 62:0-83 123:74-104
 @ ./node_modules/katex/dist/katex.css 2:26-79
 @ ./node_modules/@activeviam/plugin-widget-text-editor/dist/esm/DangerousMarkdownPreview-0f2462ba.js 5:0-30
 @ ./node_modules/@activeviam/plugin-widget-text-editor/dist/esm/index.js 19:45-93
 @ ./node_modules/@activeviam/activeui-sdk/dist/esm/index.js 81:0-79 81:0-79
 @ ./src/migrateDashboard.ts 8:23-58
 @ ./src/index.ts 5:25-54

Can you build the repo?

antoinetissier:
> Can you build the repo?

I get the same errors but it's only for fonts.
They have been here for a while and do not seem to block the build.

antoinetissier:
Thanks for the commit, it works 👍

src/migrateUIFolder.test.ts
```diff
@@ -1,16 +1,94 @@
...
+
+  it("removes the specified widget plugins from the widget bookmarks themselves, and from the content of the dashboard bookmarks", () => {
+    const keysOfWidgetPluginsToRemove = ["filters"];
+    const migratedUIFolder = migrateUIFolder(
+      legacyUIFolder,
+      servers,
+      keysOfWidgetPluginsToRemove
+    );
+
+    // In the Activeui 4 folder, the file with id `0xb` represents a saved Page Filters widget.
```
Nouzbe, line 39:
```suggestion
    // In the ActiveUI 4 folder, the file with id `0xb` represents a saved Page Filters widget.
```

src/migrateUIFolder.test.ts
```diff
@@ -1,16 +1,91 @@
...
+        widgetPluginKeysInLegacyDashboard.push(bookmark.value.containerKey)
+      )
+    );
+    const migratedDashboard = JSON.parse(
+      migratedUIFolder.children!.dashboards.children!.content.children!.eef
+        .entry.content
+    );
+    const { content, layout } = migratedDashboard.pages["p-0"];
+    const widgetPluginKeys = _map(content, ({ widgetKey }) => widgetKey);
+    expect(
```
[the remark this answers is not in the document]
Nouzbe: Don't forget this comment 🙏

src/migrateUIFolder.test.ts
```diff
@@ -1,16 +1,73 @@
...
+    // This widget is removed from the migrated dashboard in the ActiveUI 5 folder.
+    const migratedDashboard = JSON.parse(
+      migratedUIFolder.children!.dashboards.children!.content.children!.eef
+        .entry.content
+    );
+    const { content, layout } = migratedDashboard.pages["p-0"];
+    const widgetPluginKeys = _map(content, ({ widgetKey }) => widgetKey);
+    expect(
+      _intersection(widgetPluginKeys, keysOfWidgetPluginsToRemove)
+    ).toHaveLength(0);
```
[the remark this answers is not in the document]
Nouzbe: Don't forget this one either 🙏

antoinetissier:
I would like to keep the abstraction of "output plugin keys do (not) contain the plugins keys to remove", rather than projecting it on "filters".
I'll propose another way of writing with jest assertions, and if you still don't like it I'll go for your suggestion.

Nouzbe (approved):
Looks good 👍

PR #23, opened 2022-04-20
UI-7241 - Better feedback for unsupported widgets

antoinetissier:
See [UI-7241](https://support.activeviam.com/jira/browse/UI-7241).

- When migrating an unsupported saved widget: log a warning message which includes the widget name, plugin key, and content server file id
- When migrating a dashboard which includes unsupported saved widgets: log a warning message which includes the dashboard name, content server file id, and map of page key to unsupported widget names (grouped by widget plugin key)


Nouzbe:
Closed in favor of #33

PR #26, opened 2022-07-25
UI-7461 - Adds the ability to migrate calculated measures from the legacy pivot folder in AUI-4 to AUI-5

NZepeda:
## Summary
This PR introduces the ability to migrate calculated measures from AUI-4 to AUI-5. In order to do this, a new CLI parameter has been introduced `-p` which is the path to the file containing the JSON representation of the pivot folder on an AUI-4 content server.


src/cli/bin.ts
```diff
@@ -30,26 +31,43 @@ yargs
         demandOption: false,
         desc: "A list of keys of ActiveUI 4 widget plugins that should be removed during the migration.",
       });
+      args.option("pivot-input-path", {
+        alias: "p",
+        type: "string",
+        demandOption: false,
+        desc: "The path to the JSON export of the ActiveUI 4 /pivot folder",
```
antoinetissier, line 38: The `/pivot` folder is not bound only to ActiveUI 4.
We just happened to save some content in there in ActiveUI 4.
```suggestion
        desc: "The path to the JSON export of the /pivot folder on the content server",
```

src/migratePivotFolder.test.ts
```diff
@@ -0,0 +1,38 @@
...
+describe("migratePivotFolder", () => {
+  beforeAll(() => {
+    const generateId = require("./generateId");
+    let counter = 0;
+    jest.spyOn(generateId, "generateId").mockImplementation(() => {
+      const id = `00${counter}`;
+      counter += 1;
+      return id;
+    });
+  });
```
antoinetissier, line 18: This works but it doesn't feel natural to use `spyOn` when you're essentially just mocking the module.
Can't you use `jest.mock` at the root ?

src/migratePivotFolder.ts
```diff
@@ -0,0 +1,179 @@
...
+    id: string;
+  }[] = [];
+
+  if (pivotFolder?.children) {
+    await Promise.all(
+      Object.values(pivotFolder?.children).map(async ({ children = {} }) => {
+        const calculatedMeasureEntries = Object.entries(children);
+
+        const measures = await Promise.all(
+          calculatedMeasureEntries.map(async (cmEntry) => {
```
antoinetissier, line 34: If you're just going to use the value in the entry, you could just use `Object.values`, without the intermediate variable assignment.

src/migratePivotFolder.ts
```diff
@@ -0,0 +1,179 @@
...
+            const [_, entry] = cmEntry;
+            const result = await parser.parseStringPromise(entry.entry.content);
+            const id = generateId();
```
antoinetissier, line 37: For separation of concerns, I wouldn't generate the id in this function: keep the role of this function of just parsing the necessary info, and then use that info together with a generated id only outside the function.

src/migratePivotFolder.ts
```diff
@@ -0,0 +1,179 @@
...
+import xml2js from "xml2js";
+import _cloneDeep from "lodash/cloneDeep";
+import { generateId } from "./generateId";
+
+/**
+ * Extracts and parses the XML calculated measure objects from the given pivot folder.
+ */
+const getCalculatedMeasuresFromPivotFolder = async (
+  pivotFolder: ContentRecord
+): Promise<
```
antoinetissier, line 11: I think this is confusing: the calculated measures live under /pivot/entitlements/cm, not /pivot.
So either I would keep the current function name, but then implement it in a way where `pivotFolder` is really /pivot, or keep the current implementation, but adapt the function name and argument:
```suggestion
const getCalculatedMeasures = async (
  calculatedMeasuresFolder: ContentRecord
): Promise<
```

You can choose the option you prefer.

src/migratePivotFolder.ts
```diff
@@ -0,0 +1,179 @@
...
+  return calculatedMeasures;
+};
+
+/**
+ * Returns the migrated UI folder, along with the converted saved calculated measures from their location in ActiveUI 4 to their respective location in ActiveUI 5.
+ */
+export async function migratePivotFolder(
+  legacyPivotFolder: ContentRecord,
+  migratedUIFolder: ContentRecord
+) {
```
antoinetissier, line 57: I would split things differently, with more decoupling:
```
const calculatedMeasures = getCalculatedMeasures(...);
const migratedUIFolder = migrateUIFolder(..., calculatedMeasures);
```
where you would use a sub `migrateCalculatedMeasures` function in `migrateUIFolder`.

NZepeda:
> I would split things differently, with more decoupling:
> 
> ```
> const calculatedMeasures = getCalculatedMeasures(...);
> const migratedUIFolder = migrateUIFolder(..., calculatedMeasures);
> ```
> 
> where you would use a sub `migrateCalculatedMeasures` function in `migrateUIFolder`.

Hello @antoinetissier, given your above suggestion, I've added a new commit that reworks how the calculated measures are migrated. In my opinion, the new way is simpler and easier to understand for future devs. I haven't had the chance to add tests for it yet, but I'm going to hand this off to you for review and if you agree that the new way is improved, I'll go ahead and add some tests. Let me know what you think!

src/migrateCalculatedMeasure.ts
```diff
@@ -0,0 +1,29 @@
...
+      entry: {
+        isDirectory: false,
+        content: JSON.stringify({
+          expression: legacyCalculatedMeasure.expression,
+          properties: [
+            `FORMAT_STRING=${legacyCalculatedMeasure.formatStringExpression}`,
+          ],
+        }),
+        owners: ["ROLE_USER"],
+        readers: ["ROLE_USER"],
```
antoinetissier, line 25: We should keep the existing permissions on the legacy CM files, instead of arbitrarily assigning "ROLE_USER".

src/getCalculatedMeasures.ts
```diff
@@ -0,0 +1,45 @@
...
+  const calculatedMeasures: {
+    expression: string;
+    formatStringExpression: string;
+    uniqueName: string;
+  }[] = [];
+
+  if (calculatedMeasuresFolder?.children) {
+    await Promise.all(
+      Object.values(calculatedMeasuresFolder?.children).map(
+        async ({ children = {} }) => {
```
antoinetissier, line 27: I'm confused, why are there two layers of children ?

src/getCalculatedMeasures.ts
```diff
@@ -0,0 +1,45 @@
...
+          const measures = await Promise.all(
+            Object.values(children).map(async (cmEntry) => {
+              const result = await parser.parseStringPromise(
+                cmEntry.entry.content
+              );
+
+              return result.calculatedMember["$"];
```
antoinetissier, line 34: Can you add a comment to explain this line ?

src/migrateUIFolder.ts
```diff
@@ -195,6 +197,11 @@ const accumulateStructure = ({
 export function migrateUIFolder(
   legacyUIFolder: ContentRecord,
   servers: { [serverKey: string]: { dataModel: DataModel; url: string } },
+  calculatedMeasures: {
```
antoinetissier, line 200: Let's make this optional and put it at the last option to avoid breaking changes.

src/migrateUIFolder.ts
```diff
@@ -207,6 +214,18 @@ export function migrateUIFolder(
...
+    expression: string;
+    formatStringExpression: string;
+    uniqueName: string;
+    id: string;
+  }[] = calculatedMeasures
+    ? calculatedMeasures.map((measure) => ({
+        ...measure,
+        id: generateId(),
+      }))
+    : [];
```
antoinetissier, line 227: This is logic related to calculated measures, so let's handle it in `migrateCalculatedMeasures`.
It will also avoid cluttering the already big `migrateUIFolder`.

src/migrateCalculatedMeasure.ts
```diff
@@ -0,0 +1,29 @@
+interface LegacyCalculatedMeasure {
```
antoinetissier, line 1: I wouldn't put the `id` as part of this interface: it's generated on the fly, so there's no need to attach it.
Also, you should export it and reuse it as much as possible.

src/migrateCalculatedMeasures.ts
```diff
@@ -0,0 +1,68 @@
...
+    expression: string;
+    formatStringExpression: string;
+    uniqueName: string;
+    id: string;
+  }[]
+): ContentRecord {
+  return {
+    entry: {
+      isDirectory: true,
+      owners: ["ROLE_USER"],
```
antoinetissier, line 18:
```suggestion
      owners: ["ROLE_CS_ROOT"],
```

src/migrateCalculatedMeasures.ts
```diff
@@ -0,0 +1,68 @@
...
+      readers: ["ROLE_USER"],
```
antoinetissier, line 19:
```suggestion
      readers: ["ROLE_CS_ROOT"],
```

src/migrateCalculatedMeasures.ts
```diff
@@ -0,0 +1,68 @@
...
+              },
+              children: {
+                [`${legacyMeasure.id}_metadata`]: {
+                  entry: {
+                    content: JSON.stringify({
+                      name: legacyMeasure.uniqueName,
+                    }),
+                    owners: ["ROLE_USER"],
+                    readers: ["ROLE_USER"],
+                    isDirectory: false,
```
antoinetissier, line 56: Same here, we should keep the existing permissions.

src/migrateCalculatedMeasures.ts
```diff
@@ -0,0 +1,68 @@
+import { ContentRecord } from "@activeviam/activeui-sdk";
+import { migrateCalculatedMeasure } from "./migrateCalculatedMeasure";
+
+/**
+ * Returns the converted content record tree for the calculted measures, ready to be used in ActiveUI 5.
+ */
+export function migrateCalculatedMeasures(
```
antoinetissier, line 7: In terms of code split I find it a bit weird that `migrateCalculatedMeasure` just handles migrating the `content` part of the calculated measure, but that the `structure` part is inlined.

I would see it more like that:
```
function migrateCalculatedMeasures(legacyCalculatedMeasures) {
  const contentFolder = {entry: {isDirectory: true, owners: ["ROLE_USER"], readers: ["ROLE_USER"]}, children: {}}
  const structureFolder = {entry: {isDirectory: true, owners: ["ROLE_USER"], readers: ["ROLE_USER"]}, children: {}}
  legacyCalculatedMeasures.forEach(() => {
   // Add what's necessary in contentFolder
   // Add what's necessary in structureFolder
  }
  
  // return the calculated_measures directory owned by CS_ROOT, with content/structure as children
)
}
```

src/__test_resources__/uiFolderWithoutCalculatedMeasuresEntry.ts
```diff
@@ -0,0 +1,194 @@
+/**
+ * An already migrated UI folder that is missing an entry for `calculated_measures`.
+ * Useful for unit tests.
+ */
+export const uiFolderWithoutCalculatedMeasuresEntry = {
```
antoinetissier, line 5: This shouldn't be needed anymore, right ?

src/getCalculatedMeasures.ts
```diff
@@ -31,7 +22,28 @@ export const getCalculatedMeasures = async (
...
-              return result.calculatedMember["$"];
+              const { owners, readers, isDirectory, canRead, canWrite } =
+                cmEntry.entry;
+
+              const calculatedMeasure = {
+                owners,
+                readers,
+                isDirectory,
+                canRead,
+                canWrite,
```
antoinetissier, line 33: `canRead` and `canWrite` are generated on the fly when fetching a file from the content server for a given user.
It would make no sense to define it statically on the content server.
And so it's not part of the [API](https://artifacts.activeviam.com/documentation/rest/6.0.0-M1/content-server.html#files___import_true_post) when you import an entry: you just have owners and readers. 
So you just discard those attributes here.

Let me know if something is unclear.

src/migrateCalculatedMeasure.ts
```diff
@@ -1,8 +1,14 @@
-interface LegacyCalculatedMeasure {
+import { generateId } from "./generateId";
+
+export interface LegacyCalculatedMeasure {
   expression: string;
   formatStringExpression: string;
   uniqueName: string;
-  id: string;
+  canRead: boolean;
+  canWrite: boolean;
```
antoinetissier, line 8: should be removed from this interface too, and pretty much everywhere

src/migrateCalculatedMeasure.ts
```diff
@@ -11,18 +17,50 @@ interface LegacyCalculatedMeasure {
 export function migrateCalculatedMeasure(
```
antoinetissier, line 17: I still don't get why we have this indirection: as I suggested in [this snippet](https://github.com/activeviam/activeui-migration/pull/26#discussion_r932062876), I would rather inline this in the `forEach` directly in `migrateCalculatedMeasures` , because:
- it's simpler code-split wise
- it allows to do everything in one loop, rather having two first generate all the intermediate results, then go through them all

src/migrateCalculatedMeasures.ts
```diff
@@ -1,67 +1,40 @@
...
  */
 export function migrateCalculatedMeasures(
-  legacyCalculatedMeasures: {
-    expression: string;
-    formatStringExpression: string;
-    uniqueName: string;
-    id: string;
-  }[]
+  legacyCalculatedMeasures: LegacyCalculatedMeasure[],
+  calculatedMeasuresFolder: ContentRecord
```
antoinetissier, line 12: Unless I am missing something there is no need to pass `calculatedMeasuresFolder`.
You're creating an entirely new entry in the `/ui` folder with our own rules, and it shouldn't be tied structure wise to what was existing in `/pivot/entitlements/cm`.
Removing the argument will be simpler and avoid confusion.

src/getCalculatedMeasures.ts
```diff
@@ -0,0 +1,57 @@
...
+  calculatedMeasuresFolder: ContentRecord
+): Promise<LegacyCalculatedMeasure[]> => {
+  const parser = new xml2js.Parser();
+
+  const calculatedMeasures: LegacyCalculatedMeasure[] = [];
+
+  if (calculatedMeasuresFolder?.children) {
+    await Promise.all(
+      Object.values(calculatedMeasuresFolder?.children).map(
+        async ({ children = {} }) => {
```
antoinetissier, line 18: I'm still confused by this: why are there two layers of children ?
It seems to me that the children of the calculated measures folder are the calculated measures files directly.

antoinetissier (commented):
I think I wasn't clear, I think the best structure would be:
```
migrateCalculatedMeasures(legacyPivotFolder) {
  const entitlements = legacyPivotFolder?.children?.entitlements;
  const legacyCalculatedMeasuresFolder = entitlements?.children?.cm;
  const calculatedMeasures = legacyCalculatedMeasuresFolder 
    ? await getCalculatedMeasures(calculatedMeasuresFolder)
    : [];
 
const migrateCalculatedMeasuresFolder = {...}; // fill the entry
calculatedMeasures.forEach(// mutate the migrateCalculatedMeasuresFolder as required); 
}
```

The idea being that all the logic to migrate calculated measures is encapsulated in one place, and that `migrateUIFolder` itself just has to forward the right arguments.
A bit like for `migrateSettingsFolder`.
Happy to talk on the phone is something is unclear.

src/migrateCalculatedMeasures.ts
NZepeda: The only reason I passed in `calculatedMeasuresFolder` here was to get the `entry` property from it, which contains required properties such as `owners` and `readers`. Indeed each measure contains the same `owners ` and `readers` property, however, the return structure is of shape:

```
{
  entry: {
    owners: [...]
    readers: [...]
  },
  children: {
    content: {
       entry: {
        owners: [...]
        readers: [...]
       }
       // Each measure contains it's own `entry` object which also contains `owners` and `readers`
       children: {...allMeasures}
    }
  }
}
```

src/getCalculatedMeasures.ts
NZepeda: There's two levels of `children` here because the shape of the `calculatedMeasuresFolder` is as follows: 

```
{
    "entry": {...},
     "children": {
        "EquityDerivativesCube": {
           "entry": {...},
            "children": {
               // The actual measure objects live here
           }
     }
}
```

As you can see above, the first level of `children` grabs the `EquityDerivativesCube` object, but since in AUI-5, we don't distinguish the measures based on cube, I grab just the `values` with `Object.values`. I then grab the second level of `children` which actually contains the measures.

src/migrateCalculatedMeasure.ts
NZepeda: Ahh I see, I've just added a commit that inlines this directly in `migrateCalculatedMeasures`

src/migrateCalculatedMeasures.ts
```diff
@@ -1,41 +1,89 @@
...
+
+  legacyCalculatedMeasures.forEach((legacyMeasure) => {
+    const { owners, readers, expression, formatStringExpression, uniqueName } =
+      legacyMeasure;
+
+    const id = generateId();
+
+    contentFolder.children = {
+      ...contentFolder.children,
+      [id]: {
```
antoinetissier, line 45: More simply (and more performant) would be:
```
contentFolder.children[id] = ...
```
Same for `structureFolder`.

src/migrateCalculatedMeasures.ts
```diff
@@ -1,41 +1,89 @@
...
-          return { ...acc, ...currentCM.structure };
-        }, {}),
       },
+    };
+  });
+
+  return {
+    entry: {
+      ...calculatedMeasuresFolder.entry,
+    },
```
antoinetissier, line 83: Here as I mentioned before I don't think we should rely on the permissions of `/pivot/entitlements/cm`: with the ActiveUI 5 structure, we should just define a new directory entry with `ROLE_CS_ROOT` both in `readers` and `owners.`

src/migrateCalculatedMeasures.ts
```diff
@@ -1,41 +1,89 @@
...
-  calculatedMeasuresFolder: ContentRecord
-): ContentRecord {
-  const migratedCalculatedMeasures = legacyCalculatedMeasures.map(
-    (legacyMeasure) => migrateCalculatedMeasure(legacyMeasure)
-  );
+export async function migrateCalculatedMeasures(
+  legacyPivotFolder: ContentRecord
+): Promise<ContentRecord> {
+  const entitlements = legacyPivotFolder?.children?.entitlements;
+  const calculatedMeasuresFolder = entitlements?.children?.cm!;
```
antoinetissier, line 21: Are you sure that `cm` is necessarily defined in there if there are no calculated measures ?
I would handle the undefined case just out safety: you can return `{}` early in that case.

src/getCalculatedMeasures.ts
antoinetissier: Ah yes I forgot about the cubes.
Can you add a comment in the code ?
Something like:
"Legacy calculated measures are grouped by cube in `/pivot/entitlements/cm`. But they are flattened in ActiveUI 5".
And then you could use `_flatMap`.
Lastly, instead of defaulting `children` to `{}`, I would handle this case with an early return: it's more explicit so easier to read, and there is no unnecessary extra work.

src/migrateCalculatedMeasures.ts
antoinetissier: > The only reason I passed in calculatedMeasuresFolder here was to get the entry property from it, which contains required properties such as owners and readers

We don't need for the calculated measures root: the way I see it `/pivot/entitlements/cm` is managed by the server (you could have calculated measures in it that do not come from AUI4, but are just conveniently defined in XML by an admin), whereas the new `/ui/calculated_measures` is managed by AUI5.
And independently from that, the permissions on those folders depends on the underlying structure: like for any of our saved content type, we have `content` and `structure` under the root `calculated_measures`.
So the root must be strictly `ROLE_CS_ROOT` because we don't want a random user to delete it, but `calculated_measures/structure` must be writable by any user (that's what happens when you create a CM from the UI).

src/migrateCalculatedMeasures.ts
```diff
@@ -0,0 +1,89 @@
...
+    contentFolder.children = {
+      ...contentFolder.children,
+      [id]: {
+        entry: {
+          owners,
+          readers,
+          isDirectory: false,
+          content: JSON.stringify({
+            expression,
+            properties: [`FORMAT_STRING=${formatStringExpression}`],
```
antoinetissier, line 52: In the MDX spec, calculated members can have other properties than `FORMAT_STRING` (for example `MEMBER_CAPTION`).
Could you check what happens in AUI4 if you save a calculated measure with such a property ?
If it ends up in the XML, we should take it into account when migrating.

NZepeda: @antoinetissier I've just checked AUI-4, and it looks like when creating a new calculated measure, we support `FORMAT_STRING` and custom styling. I don't see how a user can save a calculated measure with the property of `MEMBER_CAPTION`. Additionally, it looks like when you save a measure in AUI-5, the only modification that can be done is to `FORMAT_STRING`, so we don't even support additional properties at the moment. Should I leave this as is? What do you think?

src/migrateCalculatedMeasures.ts
```diff
@@ -18,20 +18,20 @@ export async function migrateCalculatedMeasures(
...
+  const calculatedMeasuresFolder = entitlements?.children?.cm;
 
   const legacyCalculatedMeasures = calculatedMeasuresFolder
     ? await getCalculatedMeasures(calculatedMeasuresFolder)
     : [];
 
   const contentFolder = {
     entry: { isDirectory: true, owners: ["ROLE_USER"], readers: ["ROLE_USER"] },
-    children: {},
+    children: {} as { [id: string]: ContentRecord },
```
antoinetissier, line 29: Instead of casting here, can't you type `contentFolder` as `ContentRecord` ?
Same for `structureFolder`.

src/migrateCalculatedMeasures.ts
antoinetissier: I was thinking of trying to add the property manually in the query editor in ActiveUI 4, saving the calculated measure, and see if the property ends up on the content server.
But thinking more about it, I think the measures saved this way were actually published to the cube, and so that they were not inlined in the MDX WHERE clause.
In this case it's all good 👍

src/__test_resources__/smallLegacyPivotFolder.ts
```diff
@@ -0,0 +1,180 @@
+/**
+ * The content of the /pivot folder on a Content Server, useful for unit tests.
+ */
+export const smallLegacyPivotFolder = {
```
antoinetissier, line 4: This doesn't seem to be used anywhere right now.
Can you add a test of `migrateUIFolder` that leverages it ?

src/getCalculatedMeasures.ts
```diff
@@ -0,0 +1,55 @@
+import { ContentRecord } from "@activeviam/activeui-sdk";
+import xml2js from "xml2js";
+import _flatMap from "lodash/flatMap";
+import { LegacyCalculatedMeasure } from "./migrateCalculatedMeasures";
+
+/**
+ * Extracts and parses the XML calculated measure objects from the /pivot/entitlements/cm folder.
+ */
+export const getCalculatedMeasures = async (
```
antoinetissier, line 9: Let's add a unit test to this pure function.

antoinetissier (commented):
Almost there ! Let's just add some tests

src/migrateUIFolder.test.ts
```diff
@@ -17,20 +18,46 @@ const hasRecord = (contentRecord: ContentRecord, recordId: string): boolean =>
...
 
-  it("returns a valid ActiveUI5 /ui folder on a real life input", () => {
-    const migratedUIFolder = migrateUIFolder(legacyUIFolder, servers);
+  it("returns a valid ActiveUI5 /ui folder on a real life input", async () => {
+    const migratedUIFolder = await migrateUIFolder(legacyUIFolder, servers);
+    expect(migratedUIFolder).toMatchSnapshot();
+  });
+
+  it("returns a valid ActiveUI5 /ui folder that includes calculated measures when the input includes a pivotFolder", async () => {
+    const migratedUIFolder = await migrateUIFolder(
```
antoinetissier, line 48: I cannot see the diff on GitHub, it's too big.
This makes me think that nobody will go through the whole snapshot anyway.
So I propose to narrow down the target of the test: extract the `calculated_measures` folder from `migratedUIFolder`, and only snapshot that.

src/migrateUIFolder.test.ts
```diff
@@ -17,20 +18,46 @@ const hasRecord = (contentRecord: ContentRecord, recordId: string): boolean =>
     (child, childId) => childId === recordId || hasRecord(child, recordId)
   );
 
+jest.mock(`./generateId`, () => {
```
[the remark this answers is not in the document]
NZepeda: Unfortunately, moving the `jest.mock` under the `describe` function, does not work so I'll leave this as is.

src/migrateCalculatedMeasures.ts
```diff
@@ -40,8 +39,7 @@ export async function migrateCalculatedMeasures(
 
     const id = generateId();
 
-    // @ts-ignore `children` will always be defined since it's hardcoded as an empty object above.
-    contentFolder.children[id] = {
+    contentFolder.children![id] = {
```
antoinetissier, line 42: In my suggestion was still keeping a comment to explain why casting as non-nullable is safe.
Can you add the comment back ?
And in general, if you agree with a GitHub suggestion, could you use the "Commit suggestion" feature ? It makes both the developer and the reviewer gain time.

NZepeda: Yes definitely! In this case, I don't see the option to commit your suggestion directly so I'll a new commit manually.

antoinetissier (approved):
Looks great, just the last remark on re-adding the comments, but I am approving ✅

PR #30, opened 2022-08-03
UI-7625 - Update the repos README file

NZepeda:
## Summary
Although the dev to add the ability to migrate calculated measures has been merged into the main branch of this repo, there is no mention of this new feature in the README file. This PR addresses that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,9 @@ npm install activeui-migration
 - _content of ui.json_: the original content to migrate
 - _servers.json_: the ActivePivot server's information
 
+and a third _optional_ file:
+- _content of pivot.json_: the content from the pivot folder on the content server.
+
 Let's create these files.
 
 ### Download the original content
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -37,6 +40,16 @@ Right-click the folder named **ui** and click **Export**. This will download a f
 
 ![](/documentation/download-content.gif)
 
+### Optional: Download the content of the pivot folder
+
+:warning: If there are calculated measures to be migrated from ActiveUI-4 to the ActiveUI-5 content servers, this file is required. If this is not the case, this file can be disregarded.
+
+The pivot folder can be downloaded as a JSON file from the ActiveViam Content Manager. If you are used to accessing ActiveUI at `https://my-server.com/ui/index.html`, then you should be able to access your Content Manager at `https://my-server.com/content/ui/index.html`.
```
antoinetissier, line 47:
```suggestion
The pivot folder can be downloaded as a JSON file from the atoti+ Admin UI. If you are used to accessing ActiveUI at `https://my-server.com/ui/index.html`, then you should be able to access the Admin UI at `https://my-server.com/admin/ui/index.html`.
```

antoinetissier: I don't think we need to mention the "if you are used to..." part twice  since we mention it right above.
However, let's adapt the first one that it is mentioned with more up to date doc.

README.md
```diff
@@ -37,6 +40,16 @@ Right-click the folder named **ui** and click **Export**. This will download a f
...
+
+:warning: Make sure to connect as a user with access to all the content that must be migrated. We recommend connecting as a user with the role `ROLE_CS_ROOT`.
```
antoinetissier, line 49: Same here, I don't think we need to duplicate it.

NZepeda:
@antoinetissier Passing back to you for review but I'm getting an error that I haven't seen in the past. Have you seen this error before?

antoinetissier (approved):
One last suggestion but looks good otherwise, I am approving ✅

PR #32, opened 2022-08-15
UI-7672 - fix broken tests

rgosse:
Some things have changed in `activeui-sdk` that broke some tests in `activeui-migration`:
- `readers` of some top-level content server folders have been adjusted
- `deriveMappingFromMdx` does not map an `ALL_MEASURES` field for kpi widgets anymore
- parsing some queries can yield `slicerAxis: undefined` which is not invalid.

At the same time, tests were not run during PR checks, this also fixes that.


rgosse:
It would be best to enable the circleci project of this repo so that tests are run when commit are pushed on branches.

PR #33, opened 2022-08-15
UI-7486 - Report errors in file

rgosse:
Errors that were originally caught and printed on the console are now logged in a separate file when using the `--debug` flag.

Ex: 
```
{
  "dashboards": {
    "296": {
      "pages": {
        "p-0": {
          "pageName": "Page 1",
          "widgets": {
            "0": {
              "widgetName": "dimensions/PNLs",
              "error": {
                "message": "Cube not found: \"EquityDerivativesCube2\". Available cubes are [\"EquityDerivativesCube\",\"EquityDerivativesCubeDist\"]"
              }
            }
          }
        }
      }
    },
    "eef": {
      "pages": {
        "p-0": {
          "pageName": "Start page",
          "widgets": {
            "4": {
              "widgetName": "Page filters",
              "error": {
                "message": "Unsupported widgetKey: \"filters\""
              }
            }
          }
        }
      }
    }
  },
  "widgets": {
    "0xb": {
      "folderId": [
        "3f3"
      ],
      "folderName": [
        "BAM"
      ],
      "name": "Page filters",
      "error": {
        "message": "Unsupported widgetKey: \"filters\""
      }
    }
  }
}
```

Also adds the `--stack` flag to add stacktraces to the report.

When not in `--debug` mode, it prints a summary on the console:

```
--------- END OF CONTENT MIGRATION ---------
# Dashboards
- 15 were succesfully migrated.
- 2 were partially migrated, but errors occurred in some of the widgets they contain. These widgets were copied as is into the migrated dashboards.
- 0 could not be migrated because errors occurred during their migration. They were copied as is into the migrated folder.

# Filters
- 2 were succesfully migrated.
- 0 could not be migrated because errors occurred during their migration.

# Widgets
- 14 were succesfully migrated.
- 0 were removed.
- 1 could not be migrated because errors occurred during their migration. They were copied as is into the migrated folder.

For more information about the errors that occurred, rerun the command with the `--debug` option.
This will output a file named `report.json` containing the error messages.
To see the stack traces of the errors in this file, you can also use the `--stack` option.
--------------------------------------------
```

Some test data
https://github.com/activeviam/IBolari-AUI5-customisations/tree/mrex-ui-migration


src/migrateDashboard.ts
```diff
@@ -56,7 +59,22 @@ export function migrateDashboard(
           dashboardLeafKey,
         ];
       } else {
-        content[dashboardLeafKey] = migrateWidget(widget.bookmark, servers);
+        const [migratedWidget, widgetMigrationError] = migrateWidget(
```
Nouzbe, line 62: It would seem more conventional to have `migrateWidget` throw its error instead of returning `[whatItIsSupposedToReturn, anError?]`.

The code here would become:

```
try {
  const migratedWidget = migrateWidget()
} catch (widgetMigrationError) {
  ...
}
```

src/migration.types.ts
```diff
@@ -60,3 +60,20 @@ export interface LegacyDashboardState {
   };
   writable: boolean;
 }
+
+export type PageMigrationReport = Record<string, { warnings: string[] }>;
+
+export interface DashboardMigrationReport {
+  name: string;
+  errorStack?: string[];
+  pages?: PageMigrationReport;
```
Nouzbe, line 69: Shouldn't there be a page migration report for each page? If I am reading this correctly, there seems to be a single page migration report in here, which is a map of leaf keys (I assume) to errors.

src/migrateDashboard.ts
```diff
@@ -56,7 +59,22 @@ export function migrateDashboard(
...
+        );
+        content[dashboardLeafKey] = migratedWidget;
+        if (widgetMigrationError) {
+          if (!dashboardMigrationReport.pages) {
+            dashboardMigrationReport.pages = {};
+          }
+          if (!dashboardMigrationReport.pages[pageKey]) {
+            dashboardMigrationReport.pages[pageKey] = { warnings: [] };
+          }
+          dashboardMigrationReport.pages[pageKey].warnings.push(
```
Nouzbe, line 74: It seems misleading to push an error into an attribute named `warnings`. I would suggest to change the report type as follows:

```
interface FileErrorReport {
  // The id of each parent folder, from the root down to the direct parent.
  folderId: string[];
  // The name of each parent folder, from the root down to the direct parent.
  folderName: string[];
  // The name of the file on which the error occurred.
  name: string;
  // The thrown error.
  error: Error;
}

interface MigrationReport {
  dashboards: {
    [dashboardId]: 
      // If the error was thrown by `migrateDashboard` itself, not an underlying call to `migrateWidget`.
      // This should happen rarely.
      // In this case, the whole outdated dashboard is copied as is.
      | FileErrorReport
      // If the error was thrown by one or several underlying call(s) to `migrateWidget`.
      // This should happen more frequently.
      // In this case, the dashboard should be migrated succesfully, except for the widgets which threw errors.
      // These widgets should be copied as is into the new dashboard.
      | {
          pages: {
            [pageKey: string]:  {
              pageName: string;
              widgets: {
                [leafKey]: {
                  widgetName: string;
                  error: Error;
                }
              }
            }
          }
        }
  },
  widgets: {
    [widgetId: string]: FileErrorReport;
  },
  filters: {
    [filterId: string]: FileErrorReport;
  }
}
```

Nouzbe (commented):
I would prefer to introduce a new CLI argument allowing to control whether the report should be written. And by default, it should not. Otherwise, I am afraid that the migration might fail because of write permissions on the file system, or because the report already exists.

Alternatively, we could even only `console.log` the report and let users save their terminal output to a file themselves. Just bringing it up because it is an possible alternative, even though I would personally like it less, especially if the terminal output contains raw strings and then a JSON object (I would prefer to have the raw strings in my terminal and then the JSON object written in a JSON file for me). What do you think?

--

I would prefer to avoid repeating the following in many places of the report: 

> The widget (\"widget name\") will be copied as is. It will most likely not work correctly in ActiveUI 5. Alternatively, you can remove all widgets of this type by using the --remove-widgets option in the CLI.

Instead, I would prefer to only explain once that when errors are thrown when migrating widgets or dashboards, then these widgets or dashboards are just copied as is into the migrated folder. And that the `--remove-widgets` option exists. My motivation here is to make the report easier to use, by reducing clutter in it.

--
A few additional remarks / suggestions below:

src/migration.types.ts
Nouzbe: In other words, it seems that a dashboard report contains a map of leaf keys to errors.
Whereas I am expecting it to contain a map of page keys to a map of leaf keys to errors.

src/migrateDashboard.ts
rgosse: This function always return the body of the migrated widget, even if an error (or should we say warning) occurred. So throwing is not going to be of any help, unless we include the widget body in the thrown Error.

src/migration.types.ts
```diff
@@ -60,3 +60,20 @@ export interface LegacyDashboardState {
...
+}
+
+export interface MigrationErrorReport {
+  dashboards?: Record<string, DashboardMigrationReport>;
```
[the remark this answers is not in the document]
rgosse: Ok.

src/migration.types.ts
rgosse: There is a report for each page, it seems that the name for that type is misleading, I will inline it instead.

src/migrateDashboard.ts
rgosse: Two remarks/questions about this:
- Do you intend to put `Error` objects in the file? I'm discovering that error properties are not enumerable and therefore do not appear once serialized to json, resulting in things like:
```
"widgets": {
            "0": {
              "widgetName": "dimensions/PNLs",
              "error": {
                "cubeName": "EquityDerivativesCube2",
                "availableCubeNames": [
                  "EquityDerivativesCube",
                  "EquityDerivativesCubeDist"
                ]
              }
            }
          }
```

To get a complete vision of the error we should use another structure to include the message and the stack.
- In the comment you say that an error with the dashboard migration results in the dashboard being copied "as-is". This is not done currently, probably because we do not expect errors outside of migrating widgets within a dashboard. It is not costly to do it but I don't really see the point.

src/migrateUIFolder.ts
```diff
@@ -231,50 +257,80 @@ export async function migrateUIFolder(
...
           }
 
-          const migratedWidget = migrateWidget(bookmark, servers);
+          // Whatever the error type, an error report is created.
+          const widgetErrorReport: FileErrorReport = createFileErrorReport(
+            id,
+            bookmark.name,
+            error,
+          );
+          _setWith(migrationReport, ["widgets", id], widgetErrorReport, Object);
```
rgosse, line 330: `_set` will create an array if it can interpet any part of the path as a number, even if it is a string. Using this guarantees that only objects are created, even if `id` is a stringified number.

src/errors/UnsupportedWidgetKeyError.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Thrown when a legacy widget's key is not recognized by the migration script.
+ */
+export class UnsupportedWidgetKeyError extends Error {
+  public widgetPluginKey: string;
+  public widgetName: string;
+  constructor(widgetPluginKey: string, widgetName: string) {
+    super(
+      `Unsupported widgetKey: "${widgetPluginKey}". The widget ("${widgetName}") will be copied as is. It will most likely not work correctly in ActiveUI 5. Alternatively, you can remove all widgets of this type by using the --remove-widgets option in the CLI.`,
```
Nouzbe, line 9: Let's reduce the clutter in the report like this:

```suggestion
      `Unsupported widgetKey: "${widgetPluginKey}"
```

I would suggest to add the explanation just once in the console, at the end of the migration process if it yielded any error. For instance:

```
--------- END OF CONTENT MIGRATION ---------

- 147 dashboards were succesfully migrated.
- 59 dashboards were partially migrated, but errors occurred in some of the widgets they contain. These widgets were copied as is into the migrated dashboards.
- 3 dashboards could not be migrated because errors occurred during their migration. They were copied as is into the migrated folder.

For more information about the concerned dashboards and widgets and the errors that occurred, rerun the command with the `--debug` option. 
This will output a file named `report.json` containing the error messages.
To see the stack traces of the errors in this file, you can also use the `--stack` option.

--------------------------------------------
```

src/migration.types.ts
```diff
@@ -60,3 +60,52 @@ export interface LegacyDashboardState {
...
+        [leafKey: string]: {
+          widgetName: string;
+          error: string;
+        };
+      };
+    };
+  };
+}
+
+export interface MigrationReport {
```
Nouzbe, line 92: Since we are in the `activeui-migration` repo, I don't doubt that maintainers will understand that this object refers to a migration. The fact that it refers to errors might not be self-explanatory though, unless we express it.

```suggestion
export interface ErrorReport {
```

Also, let's add a multiline comment above it explaining what the object represents and mentioning the report file that users of the CLI can choose to output.

Same suggestion for other added types.

src/cli/bin.ts
```diff
@@ -56,16 +57,33 @@ yargs
...
+
+      if (migrationReport) {
+        await fs.writeJSON(
+          path.join(...dir, "migration-report.json"),
+          migrationReport,
+          {
+            spaces: 2,
+          },
+        );
+        console.log(
```
Nouzbe, line 81: And error log would be more adapted here.

```suggestion
        console.error(
```

Nouzbe (commented):
We're getting there. I would prefer to add new CLI options in order to control:
- whether the report is created (defaults to no)
- whether the report contains stack traces or only error messages (defaults to only error messages)

A few additional suggestions below:

src/migrateUIFolder.ts
```diff
@@ -213,7 +219,29 @@ export async function migrateUIFolder(
...
+    name: string,
+    error: {
+      message: string;
+      stack?: string[];
+    },
+  ): FileErrorReport => {
+    const legacyStructure =
+      legacyUIFolder?.children?.bookmarks?.children?.structure!;
+
+    const pathToParentFolder = _getPathToContentRecord(legacyStructure, id);
```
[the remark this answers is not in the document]
rgosse: Solution 2

Nouzbe (approved):
@rgosse I still had a few suggestions, mostly about vocabulary / documentation. I took the liberty to push some commits here in order to avoid back and forths.

I'll approve but please let me know if you'd like to challenge some of it. Also, of course please reassign if you push further changes (such as the removal of the remaining warnings you mentioned).

src/cli/bin.ts
```diff
@@ -36,36 +69,112 @@ yargs
...
+      if (
+        counters.dashboards.failed +
+          counters.dashboards.partial +
+          counters.filters.failed +
+          counters.widgets.failed >
+        0
+      ) {
+        console.log("\nRemarks:");
+        console.log(
+          "- the `text-editor` widget is not part of the plugin registry in the ActiveUI 5 starter. Make sure to add it in your project, see https://activeviam.com/activeui/documentation/latest/docs/tutorial/your-first-custom-widget#extend-activeui.",
```
Nouzbe, line 150: As mentioned above, I would prefer to not bother people who do not have a single `text-editor` widget in their content with this information.

src/migrateDashboard.ts
```diff
@@ -34,12 +37,19 @@ import { _getLegacyWidgetPluginKey } from "./_getLegacyWidgetPluginKey";
...
+  }: {
+    servers: { [serverKey: string]: { dataModel: DataModel; url: string } };
+    keysOfWidgetPluginsToRemove?: string[];
+    doesReportIncludeStacks?: boolean;
+  },
+): [DashboardState<"serialized">, DashboardErrorReport["pages"]?] {
   const pages: { [pageKey: string]: DashboardPageState<"serialized"> } = {};
   const body = legacyDashboardState.value.body;
-
+  const pagesHavingErrors: DashboardErrorReport["pages"] = {};
```
Nouzbe, line 52: If it is called `pagesSomething`, then the variable should be a `DashboardPage[]`. In this case, it is a map of errors. Why not avoiding this confusion by doing:

```suggestion
  const errorReport: Omit<DashboardErrorReport, "folderId" | "folderName"> = {};
```

This way, `migrateDashboard` would return a partial `DashboardErrorReport` instead of a `DashboardPagesErrorReport`, which I also find more logical. We could just add a comment in the signature explaining that `folderId` and `folderName` have to be added by the caller.

src/migrateUIFolder.ts
```diff
@@ -213,88 +256,215 @@ export async function migrateUIFolder(
...
-          console.error(
-            `An error occurred during the migration of filter ${id} ("${
-              bookmark.name
-              // Even though errors can be anything in theory, in practice they are always expected to be instances of Error.
-            }"). Ignoring this filter. Error:\n${(error as Error).stack}`,
-          );
+
+      // Ignoring files that do not have a matching entry in `structure`.
+      // These files are corrupted and already unreachable to the legacy application.
+      if (mapOfFolderIds[fileId] === undefined) {
```
Nouzbe, line 299: These removed entries should also appear in the error report.

Nouzbe (commented):
The commits starting from 
[Add progress bar on main loop](https://github.com/activeviam/activeui-migration/pull/33/commits/fb67eb650bc292494c3a213f995a0b866a964e6d) seem unrelated to this PR. Can you move them to one (or several separate PRs)?

rgosse:
> The commits starting from [Add progress bar on main loop](https://github.com/activeviam/activeui-migration/pull/33/commits/fb67eb650bc292494c3a213f995a0b866a964e6d) seem unrelated to this PR. Can you move them to one (or several separate PRs)?

It's related in the sense that since we've cleared the console of all sorts of feedback, a user migrating a large file finds himself staring at nothing for a while. If adding a progress bar is too much, at least consider adding a message indicating that the migration is ongoing.

Nouzbe:
> If adding a progress bar is too much, at least consider adding a message indicating that the migration is ongoing.

Don't get me wrong, I love the progress bar. I was just thinking of PR splitting. I'll review the whole thing now and see if we can merge it all.

Nouzbe (approved):
Let's go!

PR #34, opened 2022-08-17
UI-7672 - Fix and Run tests on CI

rgosse:
Some things have changed in `activeui-sdk` that broke some tests in `activeui-migration`:

- readers of some top-level content server folders have been adjusted
- `deriveMappingFromMdx` does not map an `ALL_MEASURES` field for kpi widgets anymore
- parsing some queries can yield `slicerAxis: undefined` which is not invalid.

Furthermore:
- configures circleCI  to run the tests for each commit on PR branches.
- disables linting for the time being (incorrectly references the `activeui` eslint config)


.circleci/config.yml
```diff
@@ -19,9 +19,9 @@ jobs:
       - checkout
       - login
       - run: yarn install --frozen-lockfile
-      - run: yarn build
-      - run: yarn lint
-      - run: yarn test
+      - run: yarn ci:test
+      - store_test_results:
+          path: ./reports/unit_tests/
```
antoinetissier, line 24: Is there a specific reason why you are storing the test results ? We're not doing coverage checks on `activeui-migration`.

jest-ci.config.js
```diff
@@ -0,0 +1,30 @@
...
+const path = require("path");
+const pkgDir = require("pkg-dir");
+
+const root = pkgDir.sync();
+
+if (!root) {
+  throw new Error("Could not find root.");
+}
+
+module.exports = {
```
antoinetissier, line 12: Same here, if we don't need test reports let's not bother with this setup.

antoinetissier:
> parsing some queries can yield slicerAxis: undefined which is not invalid.

Where does that come from ?
Changes to the parser/lexer are very/very rare, so I want to make sure we didn't introduce a bug.

rgosse:
> > parsing some queries can yield slicerAxis: undefined which is not invalid.
> 
> Where does that come from ? Changes to the parser/lexer are very rare, so I want to make sure we didn't introduce a bug.

https://github.com/activeviam/activeui-migration/pull/34/files#diff-009e13a7b9398e727f0ef8ad62fd79a251cc83b016b4df444b251521cd34c9a5R34

.circleci/config.yml
rgosse: I thought this was needed to see reports of failing tests on circleci. I will check how circleci behaves without it.

antoinetissier: Ah you're probably right, I'll let you double check

rgosse: I've checked and although it does not show anything anymore under the "tests" tab, maybe the console output is enough?

antoinetissier (approved):
Looks good ✅

PR #35, opened 2022-08-17
UI-7672 - Enable linting and fix some issues

rgosse:
This PR enables back the linting of the project by:
- fixing the configuration: it is impossible to extends the `activeui` config as it is not published publicly
- Uses the same ruleset as `admin-ui`
- fixes some easy issues
- disables rules that need more work to fix


antoinetissier (commented):
It's a shame that we don't have the same rule for trailing commas as in the monorepo, although I'm not sure where it comes from.
I'm approving anyway, I'll let you decide whether you want to add it or not.

antoinetissier (approved):

rgosse:
> It's a shame that we don't have the same rule for trailing commas as in the monorepo, although I'm not sure where it comes from. I'm approving anyway, I'll let you decide whether you want to add it or not.

That's weird, it should be the same rule since prettier is run through `activeui-sdk-scripts`... any idea why it is not working?

PR #37, opened 2022-08-26
UI-7486 - Update node version used for CI

rgosse:
This is to align with what we're most likely using to run tests locally. This will avoid surprises if differences of behavior between version of nodes exist, such as when raising NPEs.


Nouzbe (approved):
Looks good 👍

PR #42, opened 2022-09-19
UI-7780 - Document --stack and --debug CLI arguments

rgosse:
Adds a section to the README file to cover `--debug` and `--stack`.

Also models that `--stack` cannot be used without `--debug`.


antoinetissier:
@rgosse we have already added some documentation in the README in a bulk, so I took the liberty of just scoping this PR to this:

> Also models that --stack cannot be used without --debug.

PR #44, opened 2022-10-17
UI-0 - (UKS-5229) - remove else condition from removal key tracker

ibolari42:
The else condition was causing the widget to be removed to not be included in the migrated content. so when the content got passed later on to the `removewidget` function it was incomplete leading to an erroneous result

### DIFF
--- a/src/migrateDashboard.test.ts
+++ b/src/migrateDashboard.test.ts
@@ -278,7 +278,6 @@ describe("migrateDashboard", () => {
       servers,
       keysOfWidgetPluginsToRemove: widgetKeysToRemove,
     });
-
     const { content, layout } = dashboard.pages["p-0"];
     const widgetPluginKeys = _map(content, ({ widgetKey }) => widgetKey);
     expect(widgetPluginKeys).toEqual([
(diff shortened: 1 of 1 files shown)


Nouzbe:
[UKS-5229](https://support.activeviam.com/jira/browse/UKS-5229).

## Summary of Commerzbank issue:

### Context

- the original dashboard page contains 2 widgets:
  - a pivot table
  - a page context values

(if I understand correctly), they run:

```
npx migrate -i "content of ui.json" -o migrated-content.json -s servers.json --remove-widgets "context-values"
```

### Expectation

Only the context values widget is removed.

### Result

Both the context values and the pivot table widgets are removed.

Hi @ibolari42,

Is this understanding summarized above correct?
If it is, could you explain the fix? The `else` aims at only migrating widgets whose keys are not part of `keysOfWidgetPluginsToRemove`. In Commerzbank's case, I expect it to remove the context values widget but not the pivot table widget.

Nouzbe (commented):
Thank you for the explanation @ibolari42, it makes sense.

I would like to have a test that would have caught the issue.

After checking, we do have a test (admittedly unnecessarily complicated) testing the removal of widgets. I edited it until I found it straight forward to read.

But it does pass on latest, thus seeming to indicate that the described bug does not exist (although I don't doubt that it does).

Could you update this test so that it would have caught this bug?

PR #45, opened 2022-10-26
UI-7834 - Migrate calculated measure records

RaspberryChaos:

In AUI 5.0 calculated measures are saved to the content server. 
In 5.1 we want to publish them directly to the cube. 
Users migrating to 5.1 who have saved measures need to be able to migrate them to the cube.
<strike>
Part 1 : (This PR) Add new functions to transform a calculated measure saved on the content server into an object with the structure required in order to be added to the `entitlements/cm/` folder in ActivePivot.

Part 2: #46 - Get all calculated measures used in saved widgets and create the cm folder used in pivot/entitlements.

Part 3: Get all calculated measures used in saved dashboards.

For context, the function `createAPCalculatedMeasure` which is added here will be called by `createCMFolderWithinEntitlements`, added in the next PR.
</strike>

Updated: This PR adds a function to transform the definition of a calculated measure created with ActiveUI 5.0, into one that is natively supported by ActivePivot. See from [here](https://github.com/activeviam/activeui-migration/pull/45#issuecomment-1339419831) down.


antoinetissier (commented):
In the big picture, I would expect the code to look something like this:
```typescript
function migrateCalculatedMeasures(legacyUICalculatedMeasuresFolder: ContentRecord, pivotCalculatedMeasuresFolder: ContentRecord): ContentRecord{
  const cubeNames: {[measureName: string]: CubeName} = {};
  // for each saved widget and for each widget in each saved dashboards, parse the MDX (if applicable), extract the cube name, and assign it in `cubeNames` for each calculated measure found in the MDX.

  const migratedPivotCalculatedMeasuresFolder = produce(pivotCalculatedMeasuresFolder, () => {
      Object.entries(legacyCalculatedMeasuresFolder.children.content.children).forEach(([id, record]) => {
        const measureName = getCalculatedMeasureName(legacyCalculatedMeasuresFolder.childrrn.structure, id);
        const cubeName = cubeNames[measureName];
        if (!cubeName) {
          // add a warning in the report that the calculated measure was not migrated since it's not used anywhere
          return;
        }
        const migratedRecord = migrateCalculatedMeasure(record, measureName );
        migratedPivotCalculatedMeasuresFolder.children[cubeName] = migratedPivotCalculatedMeasuresFolder.children[cubeName] ?? defaultRecord;
        migratedPivotCalculatedMeasuresFolder.children[cubeName].children[measureName] = migratedRecord;
      })
  });

  return migratedPivotCalculatedMeasuresFolder;
}
```

If that makes sense to you, could you adapt the functions that you are defining in this first PR ?
If something is unclear, let me know and we can discuss this live.

RaspberryChaos:
To simplify things, this PR now only adds the functions `getCalculatedMeasureIds` and `getCalculatedMeasureNames`, plus unit tests for both these functions and test resources necessary for these tests.

The return value of `getCalculatedMeasureNames` will be passed as the argument `namesOfCalculatedMeasuresToMigrate` to the functions `migrateCalculatedMeasuresInWidgets` (#49) and `migrateCalculatedMeasuresInDashboards` (#52).

For context, I plan to follow the structure detailed [above](https://github.com/activeviam/activeui-migration/pull/45#pullrequestreview-1159658133) in the next pull request and create the `migrateCalculatedMeasures` function where these functions will be used.

antoinetissier:
Indeed we need to make a first pass to gather the names of the calculated measures to migrate. So the snippet above would be adapted like this:
```typescript
function migrateCalculatedMeasures(contentServer: ContentRecord): void {
  const legacyCalculatedMeasuresFolder = contentServer.children.ui.children.calculated_measures;
  const pivotFolder = contentServer.children.pivot;

  const legacyCalculatedMeasureRecords: {[measureName: string]: ContentRecord} = _mapKeys(legacyCalculatedMeasuresFolder.children.content.children, (id) => getCalculatedMeasureName(legacyCalculatedMeasuresFolder.children.structure, id));
  
  const namesOfCalculatedMeasuresToMigrate = Object.keys(legacyCalculatedMeasureRecords);

  const measureToCubeMappingInWidgets = migrateCalculatedMeasuresInWidgets();
  const measureToCubeMappingInDashboards = migrateCalculatedMeasuresInDashboards();
  const measureToCubeMapping = "TODO"; // merge the mapping from widgets and from dashboards  

      Object.entries(legacyCalculatedMeasureRecords).forEach(([measureName, record]) => {
        const cubeNames = measureToCubeMapping [measureName];
        if (!cubeNames) {
          // add a warning in the report that the calculated measure was not migrated since it's not used anywhere
          return;
        }
        
        const migratedRecord = migrateCalculatedMeasure(record, measureName);

        cubeNames.forEach((cubeName) => {
            pivotFolder.children[cubeName] = pivotFolder.children[cubeName] ?? defaultRecord;
            pivotFolder.children[cubeName].children[measureName] = migratedRecord;
        });
      })
  });

  delete contentServer.children.ui.children.calculated_measures;
}
```

RaspberryChaos:
This PR has been updated (several times 😩 !), so now only the last commit is relevant. As part of creating the `migrateCalculatedMeasures` function with the structure suggested by @antoinetissier [here](https://github.com/activeviam/activeui-migration/pull/45#issuecomment-1339419831), this PR adds a function `migrateCalculatedMeasureRecord`. It will be called by the `migrateCalculatedMeasures` function.

This corresponds to the line 
`const migratedRecord = migrateCalculatedMeasure(record, measureName);`
in the example above.

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasureRecord.ts
```diff
@@ -0,0 +1,34 @@
...
+  calculatedMeasureName: string,
+): ContentRecord => {
+  const migratedRecord = produce(legacyCalculatedMeasureContent, (draft) => {
+    const parsedContent = JSON.parse(draft.entry.content);
+
+    const expression = parsedContent.expression;
+    const properties: string[] = parsedContent.properties;
+
+    const formatStringExpression = properties
+      ? properties.find((property) => property.startsWith("FORMAT_STRING"))
```
RaspberryChaos, line 18: In response to [this comment](https://github.com/activeviam/activeui-migration/pull/45#discussion_r1007975828) on the earlier version of this function
> Is there a reason why it's `startsWith` and not a strict equality check `===` ?

`properties` is an array of strings, for example `FORMAT_STRING = "#,###.##"`

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasureRecord.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const migrateCalculatedMeasureRecord = (
+  legacyCalculatedMeasureContent: ContentRecord,
+  calculatedMeasureName: string,
+): ContentRecord => {
+  const migratedRecord = produce(legacyCalculatedMeasureContent, (draft) => {
+    const parsedContent = JSON.parse(draft.entry.content);
+
+    const expression = parsedContent.expression;
+    const properties: string[] = parsedContent.properties;
```
antoinetissier, line 15:
```suggestion
    const {expression, properties} = JSON.parse(draft.entry.content);
```

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasureRecord.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+    `);
+
+    expect(JSON.parse(migratedRecord.entry.content)).toStrictEqual({
+      additionalProperties: {},
+      className:
+        "com.quartetfs.biz.pivot.definitions.impl.CalculatedMemberDescription",
+      expression:
+        "Count(Descendants([Geography].[City].CurrentMember, [Geography].[City].[City]), EXCLUDEEMPTY)",
+      formatStringExpression: `FORMAT_STRING = "#,###.##"`,
```
antoinetissier, line 38: This is not the expected format.
See this example from the ranch:
image

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasureRecord.ts
```diff
@@ -0,0 +1,34 @@
...
+    const formatStringExpression = properties
+      ? properties.find((property) => property.startsWith("FORMAT_STRING"))
+      : undefined;
+
+    const migratedCalculatedMeasureContent = JSON.stringify({
+      // The className comes from Active Pivot.
+      // See https://github.com/activeviam/activepivot/blob/876981bef9a65acbb228f97c53825a356de59382/pivot/core/impl/src/main/java/com/quartetfs/biz/pivot/definitions/impl/CalculatedMemberDescription.java
+      className:
+        "com.quartetfs.biz.pivot.definitions.impl.CalculatedMemberDescription",
+      additionalProperties: {},
```
antoinetissier, line 26: We should preserve the other properties, as a map from property name to property expression.

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasureRecord.ts
```diff
@@ -0,0 +1,58 @@
...
+    expression,
+    properties,
+  }: { expression: string; properties: string[] } = JSON.parse(
+    legacyCalculatedMeasureContent.entry.content,
+  );
+
+  const formatStringExpression = properties
+    ? properties
+        .find((property) => property.startsWith("FORMAT_STRING"))
+        ?.replace("FORMAT_STRING = ", "")
```
RaspberryChaos, line 27: Is this OK like this, or do you prefer that I user `parse` here, like below ?

antoinetissier: Now that I am seeing this, I am thinking that indeed `parse` is not the safest, since it relies on the assumption that the expression of a property is necessarily an `MdxLiteral`.

So instead we could just do `const [propertyName, propertyExpression] = propertyString.split(" = ")`.

And we could do the same handling both for `"FORMAT_STRING"` and other prroperties.

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasureRecord.test.ts
```diff
@@ -0,0 +1,99 @@
+import { migrateCalculatedMeasureRecord } from "./migrateCalculatedMeasureRecord";
+import _cloneDeep from "lodash/cloneDeep";
+
+const distinctCountCity = {
+  entry: {
+    content:
+      '{"expression":"Count(Descendants([Geography].[City].CurrentMember, [Geography].[City].[City]), EXCLUDEEMPTY)","properties":["FORMAT_STRING = \\"#,###.##\\"","CAPTION = \\"Distinct count city\\" "]}',
```
antoinetissier, line 7: Did you copy this from the content server ? I am surprised that there is a trailing whitespace after 'Distinct count city'.

RaspberryChaos: No, I didn't copy it from the content server, none of the calculated measures saved on the content server had any additional properties so I just made it up. It's my mistake, I made a typo, I will fix it !

RaspberryChaos:
As discussed @antoinetissier, I have refactored this. I wrote both versions and decided to go with `migrateCalculatedMeasureContent` instead of `migrateCalculatedMeasureRecord`. I feel that it is slightly simpler and that it makes the tests more readable. You can see the `migrateCalculatedMeasureRecord` version in the previous commit if you would like to compare them, and I can revert the latest commit if you prefer the other version, let me know.

antoinetissier (commented):
Looks great ! Just one last thing: if we do it this way (= scoping the function to migrating the content as opposed to the whole record) then I think it's a shame to return the stringified version.
It would be easier to test, and more consistent to have this signature `migrateCalculatedMeasureContent(legacyParsedContent) => newParsedContent`

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasureContent.ts
```diff
@@ -4,7 +4,13 @@
 export const migrateCalculatedMeasureContent = (
   legacyCalculatedMeasureContent: { expression: string; properties: string[] },
   calculatedMeasureName: string,
-): string => {
+): {
+  className: string;
+  additionalProperties: { [propertyName: string]: string };
+  uniqueName: string;
+  expression: string;
+  formatStringExpression: string | undefined;
```
antoinetissier, line 12:
```suggestion
  formatStringExpression?: string;
```

antoinetissier (approved):
🚀

PR #47, opened 2022-11-01
UI-0 - (UKS-5248) handle unexpected values in filter objects

ibolari42:
This PR adds a function that checks the filters provided in each dashboard for non-string elements and removes then  as the only expected elements in the filter are mdx strings.

### DIFF
--- a/src/migrateDashboard.ts
+++ b/src/migrateDashboard.ts
@@ -61,11 +61,10 @@ function addWidgetErrorToReport(
   };
 }
 
-/*
-    ensure that unexpected elements in the dashboard and page filter objects are removed, this will prevent avoidable migration failures
-    Also keep track of what dashboards and pages have had this treatment
-    It is expected that this treatment will have no negative impact because the unexpected elements would have been ignored by AUI4 anyway
-    but keep track for customer's peace of mind
+/**
+   * Ensure that unexpected elements in the dashboard and page filter objects are removed, to prevent avoidable migration failures.
+   * This can occur when filters have not been serialized correctly while saving content in AUI4.
+   * This has no functional impact on the corresponding dashboards, since those filters would be ignored by AUI4 anyway.
   */
 const sanitiseFilters = (
   filters: any[],


src/migrateDashboard.ts
```diff
@@ -61,6 +61,61 @@ function addWidgetErrorToReport(
...
+        widgets: {},
+      };
+    }
+
+    if (doesReportIncludeStacks) error.stack = rejectedFilters;
+
+    errorReport.pages[sectionKey].widgets[`All widgets in ${pageName}`] = {
+      widgetName: "All widgets",
+      error,
+    };
```
antoinetissier, line 114: This is a bit of a workaround.
I would prefer to introduce an `error` at the root of `DashboardErrorReport` and at the root of each page if necessary, in a preliminary PR.

src/migrateDashboard.ts
```diff
@@ -61,6 +61,61 @@ function addWidgetErrorToReport(
...
+  filters: any[],
+  errorReport: DashboardErrorReport,
+  {
+    doesReportIncludeStacks,
+    pageKey,
+    pageName,
+  }: {
+    doesReportIncludeStacks?: boolean;
+    pageKey: string;
+    pageName: string;
```
antoinetissier, line 80: Following the same idea, I would make those optional, as it is not needed for dashboard filters.

src/migrateDashboard.ts
```diff
@@ -61,6 +61,61 @@ function addWidgetErrorToReport(
   };
 }
 
+/*
+    ensure that unexpected elements in the dashboard and page filter objects are removed, this will prevent avoidable migration failures
+    Also keep track of what dashboards and pages have had this treatment
+    It is expected that this treatment will have no negative impact because the unexpected elements would have been ignored by AUI4 anyway
+    but keep track for customer's peace of mind
+  */
```
antoinetissier, line 69: We have a few conventions for code comments:
- one line per sentence
- each sentence starts with a capital letter and ends with a dot

```suggestion
/**
   * Ensure that unexpected elements in the dashboard and page filter objects are removed, to prevent avoidable migration failures.
   * This can occur when filters have not been serialized correctly while saving content in AUI4.
   * This has no functional impact on the corresponding dashboards, since those filters would be ignored by AUI4 anyway.
  */
```

antoinetissier (commented):
Thanks for the PR @ibolari42.

Besides the comments below, it seems that the two commits are orthogonal. Could you separate PRs for each of them to keep changes atomic ?

PR #48, opened 2022-11-07
UI-7834 - Add 5.1 SDK dependency

RaspberryChaos:
Part of [UI-7834,](https://support.activeviam.com/jira/browse/UI-7834) Migrate calculated measures from the content server to the AP cube. 

Bump the version of the SDK in `package.json` in order to use the function `removeCalculatedMemberDefinition`, which was exported in [PR #2957](https://github.com/activeviam/activeui/pull/2957).

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "@activeviam/activeui-sdk-5.0": "npm:@activeviam/activeui-sdk@5.0.15",
+    "@activeviam/activeui-sdk-5.1": "npm:@activeviam/activeui-sdk@5.1.0-20221018171954-6275d65",
     "@activeviam/content-server-initialization-5.0": "npm:@activeviam/content-server-initialization@5.0.15",
     "@activeviam/data-model-5.0": "npm:@activeviam/data-model@5.0.15",
     "@activeviam/mdx-5.0": "npm:@activeviam/mdx@5.0.15",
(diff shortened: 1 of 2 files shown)


src/migrateSettingsFolder.ts
```diff
@@ -21,7 +21,9 @@ const emptySettingsFolders = _pick(
 function migrateSettingsMap(legacySettingsMap: {
   [settingKey: string]: any;
 }): Partial<Settings> {
-  const migratedSettingsMap: Partial<Settings> = {};
+  const migratedSettingsMap: Partial<
+    Settings & { "userFilters.areEnabled"?: boolean }
```
RaspberryChaos, line 25: `userFilters.areEnabled` has been replaced by the `canUseUserFilters` permission in 5.10.
It was deprecated [here](https://github.com/activeviam/activeui/pull/2653) and removed from the `Settings` type [here](https://github.com/activeviam/activeui/pull/2685).

antoinetissier: We'll need to import types from different versions of the SDK to be precise: here, it should be the settings of 5.0.
It seems like we should be able to do it with aliases: https://dev.to/3imed_jaberi/how-to-install-multiple-versions-of-a-npm-package-at-the-same-time-project-2a29.
Could you try it out ?

antoinetissier (approved):
🚀

PR #49, opened 2022-11-07
UI-7834 - Remove calculated measure definitions from widgets

RaspberryChaos:
Jira ticket: [UI-7834](https://support.activeviam.com/jira/browse/UI-7834) subtask of [UI-7822](https://support.activeviam.com/jira/browse/UI-7822).

As part of migrating calculated measures from the content server to the AP cube, this PR adds a function to remove the calculated measure definitions from the widgets in the `ui/widgets` folder and to get the `cubeName` for each calculated measure to be migrated.

For context, the intention is to use these new functions in a `migrateCalculatedMeasures` function, the structure of which is described [here.](https://github.com/activeviam/activeui-migration/pull/45#pullrequestreview-1159658133)


RaspberryChaos:
This PR is currently failing the circleci checks because it is based off another branch where we bump the version of the SDK, I'm working on a fix in #48

src/migrateCalculatedMeasuresToActivePivot/removeCalculatedMemberDefinitionFromMDXAndGetCubeName.ts
```diff
@@ -0,0 +1,48 @@
...
+): {
+  calculatedMeasuresWithCubeNames: { [measureName: string]: CubeName };
+  stringifiedUpdatedMdx: MdxString;
+} => {
+  const calculatedMeasuresWithCubeNames: { [measureName: string]: CubeName } =
+    {};
+  const parsedMdx: MdxSelect = parse(mdx);
+
+  const calculatedMeasuresKeysUsedByWidgetToBeMigrated = _intersection(
+    // The keys are the names of the calculated measures, a widget may have more than one calculated measure.
```
antoinetissier, line 26: I don't think we need to specify that a widget may have several calculated measures in its query.
It's the MDX spec, and the function name `getCalculatedMeasures` is plural anyway.
About the variable name, I would align it on the function argument.
```suggestion
  const namesOfCalculatedMeasuresToMigrateInWidget = _intersection(
```

src/migrateCalculatedMeasuresToActivePivot/removeCalculatedMemberDefinitionFromMDXAndGetCubeName.ts
```diff
@@ -0,0 +1,48 @@
...
+  MdxSelect,
+  getCalculatedMeasures,
+  getCubeName,
+  removeCalculatedMemberDefinition,
+  stringify,
+} from "@activeviam/mdx";
+import { parse } from "@activeviam/activeui-sdk";
+import _intersection from "lodash/intersection";
+
+export const removeCalculatedMemberDefinitionFromMDXAndGetCubeName = (
```
antoinetissier, line 13: The current name is pretty heavy.
I would make it simpler, and add a multiline comment to explain what it does.
```suggestion
export const migrateCalculatedMeasuresInMdx = (
```

src/migrateCalculatedMeasuresToActivePivot/removeCalculatedMemberDefinitionFromMDXAndGetCubeName.ts
```diff
@@ -0,0 +1,48 @@
...
+  mdx: MdxString,
+  namesOfCalculatedMeasurestoMigrate: string[],
+  dataModel: DataModel,
+): {
+  calculatedMeasuresWithCubeNames: { [measureName: string]: CubeName };
```
antoinetissier, line 18: This makes sense to have this structure accross several MDX queries (= several widgets), but here for a single MDX string there is only one cube name.
So the return type could be instead:
`{cubeName: CubeName, namesOfMigratedCalculatedMeasures: string[], migratedMdx: MdxString}`
And the variable names and logic can be adapted accordingly.

src/migrateCalculatedMeasuresToActivePivot/removeCalculatedMemberDefinitionFromMDXAndGetCubeName.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  "activeui5 calculated measure",
+  "Test calculated measure",
+  "pvSum ^ 2",
+];
+
+const dataModel = dataModelsForTests.sandbox;
+
+describe("removeCalculatedMemberDefinitionFromMDXAndGetCubeName", () => {
+  it("returns `calculatedMeasuresWithCubeNames` as an empty object and does not modify the MDX when there are no calculated measures", () => {
+    const removeCalculatedMeasures =
```
antoinetissier, line 27: Lets not use a verb-like name for a variable that is not a function.
You could just destructure the function output:
`const {cubeName, namesOfMigratedCalculatedMeasures, migratedMdx} = migrateCalculatedMeasuresInMdx(...)`.

Same for all tests.

antoinetissier (commented):
It would be nice to have a test that checks that a calculated measure definition remains untouched if the name does not match any of the `namesOfCalculatedMeasuresToMigrate`.

src/__test_resources__/aui5.0LegacyTestResources/widgetsFolderWithCalculatedMeasuresFrom2Cubes.ts
```diff
@@ -0,0 +1,117 @@
+export const widgetsWithCalculatedMeasuresFrom2Cubes = {
```
antoinetissier, line 1: It doesn't look like these two test files are being used yet.
If so, let's not add them in this PR.

src/migrateCalculatedMeasuresToActivePivot/removeCalculatedMemberDefinitionFromMDXAndGetCubeName.ts
RaspberryChaos: I agree the name was bad, this is definitely better.

src/__test_resources__/aui5.0LegacyTestResources/widgetsFolderWithCalculatedMeasuresFrom2Cubes.ts
RaspberryChaos: They were being used in `migratedCalculatedMeasuresInWidgets.ts`, I have now decided to combine both into one file instead.

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const migrateCalculatedMeasuresInWidgets = (
+  widgets: ContentRecord,
+  dataModel: DataModel,
+  namesOfCalculatedMeasurestoMigrate: string[],
+): {
+  cubeNames: { [measureName: string]: CubeName };
+  migratedWidgetsRecord: ContentRecord;
+} => {
+  // Create an empty object where each calculated measure used in a saved widget will be added as a key with its cubeName as a value.
```
antoinetissier, line 21: Just to avoid the confusion of "saved widgets" possibly refering to only standalone widgets, as opposed to widgets in saved dashboards:
```suggestion
  // Create an empty object where each calculated measure used in a widget will be added as a key with its cubeName as a value.
```

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,62 @@
...
+  DataModel,
+  MdxSelect,
+  WidgetWithQueryState,
+} from "@activeviam/activeui-sdk";
+import { MdxString } from "@activeviam/mdx";
+import { migrateCalculatedMeasuresInMdx } from "./migrateCalculatedMeasuresInMdx";
+import { produce } from "immer";
+import _mapValues from "lodash/mapValues";
+
+export const migrateCalculatedMeasuresInWidgets = (
```
antoinetissier, line 13: Let's explain roughly what we're doing:
```suggestion
/**
* In ActiveUI 5.0, saved calculated measures live under /ui/calculated_measures, and ActivePivot is not aware of them. In particular, they don't appear in the data model, and they must be added as query-scoped measures to each individual widget refering to them.
* 
* In ActiveUI 5.1, saved calculated measures are grouped by cube and live under /pivot/entitlements/cm. ActivePivot proxies them and makes them accessible in the data model. So they can be referred to just like any other measure in an MDX query, without the need to add a query-scoped calculated measure definition.
*/
export const migrateCalculatedMeasuresInWidgets = (
```

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,62 @@
...
+  // Create an empty object where each calculated measure used in a saved widget will be added as a key with its cubeName as a value.
+  const cubeNames: { [measureName: string]: CubeName } = {};
+
+  const migratedWidgetsRecord = produce(widgets, (draft) => {
+    draft.children!.content.children = _mapValues(
+      widgets.children!.content.children,
+      (widgetRecord) => {
+        const widgetState: WidgetWithQueryState<MdxSelect, "serialized"> =
+          JSON.parse(widgetRecord.entry.content);
+        const mdx: MdxString | undefined = widgetState.query.mdx;
```
antoinetissier, line 30: What you're doing with parsing/stringifying is perfectly correct, but now I am thinking that it would be nice to use the predefined `deserializeWidgetState`, `serializeWidgetState` instead: it does some of the work for us, and it allows to call `deserializeWidgetState` from `5.0` and `serializeWidgetState` from 5.1.

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+describe("migrateCalculatedMeasuresInWidgets", () => {
+  const { cubeNames, migratedWidgetsRecord } =
+    migrateCalculatedMeasuresInWidgets(
+      uiWidgetsFolder,
+      dataModel,
+      namesOfCalculatedMeasurestoMigrate,
+    );
+
+  it("returns a `cubeNames` object containing the names of all calculated measures used in a `ui/widgets` folder with their corresponding cube name", () => {
```
antoinetissier, line 24:
```suggestion
  it("returns a `cubeNames` object containing the names of all calculated measures used in the `ui/widgets` and `ui/dashboards` folders with their corresponding cube name", () => {
```

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    expect(cubeNames).toStrictEqual({
+      "Distinct count city": "EquityDerivativesCube",
+      "EXP pnl.Forex": "EquityDerivativesCube",
+      "Log pv.SUM": "EquityDerivativesCube",
+      "Test calculated measure": "EquityDerivativesCube",
+      "pvSum ^ 2": "EquityDerivativesCubeDist",
+    });
+  });
+
+  it("returns the widget record unchanged if a widget at the root contains no calculated measures", () => {
```
antoinetissier, line 34: Nitpick but what is returned is the `widgets` folder as a whole, not one individual widget record:
```suggestion
  it("does not change a widget record if there is no calculated measures in its MDX query", () => {
```

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    // "7a5" is a widget containing no calculated measures at the root. This is definitely defined, otherwise the snapshot would be empty.
```
antoinetissier, line 35: This is good to justify using `!` to cast as non-nullable in general, but in tests we can be a bit more lenient.
Here in particular, I wouldn't want to give the impression that it's linked to the sentence above.
So either I would move that second sentence below - as close as possible to what it refers to, or just remove it.
Same for other occurrences in the file.

Independently from that, I don't understand "otherwise the snapshot would be empty" => we're not using a snapshot here ?

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    expect(
+      migratedWidgetsRecord.children?.content.children!["7a5"],
+    ).toStrictEqual(uiWidgetsFolder.children?.content.children!["7a5"]);
+  });
+
+  it("does not remove a calculated measure definition from the MDX if it is not on the list of `namesOfCalculatedMeasuresToMigrate`", () => {
```
antoinetissier, line 41: 👍

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    // "ee7" is a widget containing 1 calculated measure which is not on the list of `namesOfCalculatedMeasuresToMigrate`. This is definitely defined, otherwise the snapshot would be empty.
+    expect(
+      migratedWidgetsRecord.children?.content.children!["ee7"],
+    ).toStrictEqual(uiWidgetsFolder.children?.content.children!["ee7"]);
+  });
+
+  it("removes the calculated measure definition from a widget containing 1 calculated measure at the root", () => {
```
antoinetissier, line 48: What does "at the root" mean here ? Same for other occurrences.
I would suggest to remove it, but if I'm missing something let me know.
```suggestion
  it("removes the calculated measure definition from a widget containing 1 calculated measure", () => {
```

EDIT: after reading below I think you mean "at the root" vs "in a folder", right ?
If so, I still think it's okay remove "at the root" and to just specify "widget in a folder" where needed.
Because if we keep the current wording I think most people would think 
 that "at the root" refers to the calculated measure, and wonder what that means.

antoinetissier (commented):
Some minor suggestions but it looks great !
In particular, really appreciate the comments in the tests. It would be pretty obscure without them, and they really make the tests easier to read 👍

antoinetissier (commented):
Just to be sure we're on the same page: we'll also need to migrate calculated measures in widgets in saved dashboards under `/ui/dashboards`.
But we can do that in a separate PR if you prefer.

src/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.test.ts
RaspberryChaos: In that case I will just remove the justifications for using `!`. 
It's true that we are not using a snapshot here, that's my mistake, I just copy/pasted "otherwise the snapshot would be empty" from one of the other tests where I used `.toMatchInlineSnapshot()`, but I'm removing this now anyway.

RaspberryChaos:
I have updated the folder structure to be inline with the changes made in #54 

I have created a 5.0_to_5.1 folder and inside it I have created a `migrateCalculatedMeasuresToActivePivot` folder to regroup all the functions related to migrating calculated measures.

folderStructure

RaspberryChaos:
> Just to be sure we're on the same page: we'll also need to migrate calculated measures in widgets in saved dashboards under `/ui/dashboards`. But we can do that in a separate PR if you prefer.

In progress here #52

src/5.0_to_5.1/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,82 @@
...
+): {
+  cubeNames: { [measureName: string]: CubeName };
+  migratedWidgetsRecord: ContentRecord;
+} => {
+  // Create an empty object where each calculated measure used in a saved widget will be added as a key with its cubeName as a value.
+  const cubeNames: { [measureName: string]: CubeName } = {};
+
+  const migratedWidgetsRecord = produce(widgets, (draft) => {
+    draft.children!.content.children = _mapValues(
+      widgets.children!.content.children,
```
antoinetissier, line 33: Just to make sure there's no undesired mutation:
```suggestion
      draft.children!.content.children,
```

src/5.0_to_5.1/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,82 @@
...
+export const migrateCalculatedMeasuresInWidgets = (
+  widgets: ContentRecord,
+  dataModel: DataModel,
+  namesOfCalculatedMeasurestoMigrate: string[],
+): {
+  cubeNames: { [measureName: string]: CubeName };
+  migratedWidgetsRecord: ContentRecord;
+} => {
+  // Create an empty object where each calculated measure used in a saved widget will be added as a key with its cubeName as a value.
+  const cubeNames: { [measureName: string]: CubeName } = {};
```
antoinetissier, line 29: What do you think of the following name instead ?
```suggestion
  const measureToCubeMapping: { [measureName: string]: CubeName } = {};
```

src/5.0_to_5.1/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInMdx.ts
```diff
@@ -0,0 +1,44 @@
...
+  getCalculatedMeasures,
+  getCubeName,
+} from "@activeviam/mdx-5.0";
+import { removeCalculatedMemberDefinition } from "@activeviam/mdx-5.1";
+import _intersection from "lodash/intersection";
+
+/**
+ * Removes any calculated measures definitions from an MDX string matching the `namesOfCalculatedMeasuresToMigrate`.
+ * Returns the updated MDX string, the names of calculated measures to be migrated from that particular widget and the corresponding cube name.
+ */
```
antoinetissier, line 13:
```suggestion
/**
 * Removes the definitions of the calculated measures matching `namesOfCalculatedMeasuresToMigrate` from the `WITH` clause of `mdx`.
 * Returns the updated MDX, the names of those calculated measures, and the corresponding cube name.
 */
```

src/5.0_to_5.1/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInMdx.test.ts
```diff
@@ -0,0 +1,111 @@
+import { dataModelsForTests } from "@activeviam/data-model-5.0";
+import { MdxSelect, parse, stringify } from "@activeviam/mdx-5.0";
+import { migrateCalculatedMeasuresInMdx } from "./migrateCalculatedMeasuresInMdx";
+
+const mdxSelectWithNoCalculatedMeasures: MdxSelect = parse(
```
antoinetissier, line 5: It would be nice to use the indented version of the MDX so that it's easier to read.
Same for the test snapshots, we could use `stringify(mdx, {indent: true})`.
If it's too much of a hassle no worries though, it's not the biggest priority.

antoinetissier (commented):
The folder name could just be `calculated-measures`

src/5.0_to_5.1/migrateCalculatedMeasuresToActivePivot/migrateCalculatedMeasuresInWidgets.ts
RaspberryChaos: I agree this is better

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInMdx.test.ts
```diff
@@ -0,0 +1,141 @@
+import { dataModelsForTests } from "@activeviam/data-model-5.0";
+import { MdxSelect, parse, stringify } from "@activeviam/mdx-5.0";
+import { migrateCalculatedMeasuresInMdx } from "./migrateCalculatedMeasuresInMdx";
+
+const mdxSelectWithNoCalculatedMeasures: MdxSelect = parse(
+  "SELECT NON EMPTY Hierarchize(Descendants({[Geography].[City].[AllMember]}, 1, SELF_AND_BEFORE)) ON ROWS, NON EMPTY {[Measures].[contributors.COUNT]} ON COLUMNS FROM [EquityDerivativesCube] CELL PROPERTIES VALUE, FORMATTED_VALUE, BACK_COLOR, FORE_COLOR, FONT_FLAGS",
```
antoinetissier, line 6: I meant that it would be nice to have the indented version here too, which is what I was thought could be more of a hassle 😅

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInMdx.ts
```diff
@@ -0,0 +1,44 @@
...
+  const cubeName = getCubeName(mdx);
+  let updatedMdx: MdxSelect = mdx;
+  namesOfCalculatedMeasuresToMigrateInWidget.forEach((calculatedMeasure) => {
+    updatedMdx = removeCalculatedMemberDefinition(updatedMdx, {
+      dimensionName: "Measures",
+      hierarchyName: "Measures",
+      namePath: [calculatedMeasure],
+      cube: getCube(dataModel, cubeName),
+    });
+  });
```
antoinetissier, line 37: This is pefectly fine, but this is typically the use case to use `reduce` rather than `forEach` (iterate several times over the same object, and use the result of one iteration in the other).

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,82 @@
...
+            cubeName,
+          } = migrateCalculatedMeasuresInMdx(
+            mdx,
+            namesOfCalculatedMeasurestoMigrate,
+            dataModel,
+          );
+
+          namesOfCalculatedMeasuresToMigrateInWidget.forEach(
+            (calculatedMeasureName) => {
+              measureToCubeMapping[calculatedMeasureName] = cubeName;
```
antoinetissier, line 61: Actually I am thinking that instead of `{[measureName: string]: string}` we should have `{[measureName: string]: string[]}`.
I.e. we should have the possibility for a calculated measure to be in several cubes.

Indeed, the following was possible for example in ActiveUI 5.0:
- create a calculated measure from a widget targeting CubeA
- save that measure
- add the save measure in a widget targeting CubeB

This is a bit of an edge case, but if the calculated measure expression makes sense in both cubes then we should support imo.

Concretely, here it would just be:
```suggestion
              measureToCubeMapping[calculatedMeasureName] =  measureToCubeMapping[calculatedMeasureName] ? _uniq([...measureToCubeMapping[calculatedMeasureName], cubeName]) : [cubeName];
```

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInMdx.test.ts
RaspberryChaos: Oops, sorry 😆. I've indented this now too

antoinetissier (approved):
Let's go 🚀

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+describe("migrateCalculatedMeasuresInWidgets", () => {
+  const { measureToCubeMapping, migratedWidgetsRecord } =
+    migrateCalculatedMeasuresInWidgets(
+      uiWidgetsFolder,
+      dataModel,
+      namesOfCalculatedMeasurestoMigrate,
+    );
+
+  it("returns a `measureToCubeMapping` object containing the names of all calculated measures used in the `ui/widgets` and `ui/dashboards` folders with their corresponding cube name", () => {
```
antoinetissier, line 27:
```suggestion
  it("returns a `measureToCubeMapping` object containing the names of all calculated measures used in the `ui/widgets` folder with their corresponding cube name", () => {
```

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,89 @@
...
+                    ])
+                  : [cubeName];
+            },
+          );
+          const updatedWidgetState = produce(
+            deserializedWidgetState,
+            (draft) => {
+              draft.query.mdx = migratedMdx;
+            },
+          );
```
antoinetissier, line 76: No need to use `produce` here since there is already one above:
```suggestion
        deserializedWidgetState.query.mdx = migratedMdx;
```

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -0,0 +1,89 @@
...
+          const updatedRecord = produce(widgetRecord, (draft) => {
+            draft.entry.content = JSON.stringify(
+              serializeWidgetState(updatedWidgetState),
+            );
+          });
```
antoinetissier, line 81: Same here
```suggestion
          widgetRecord.entry.content = JSON.stringify(
              serializeWidgetState(updatedWidgetState),
            );
```

antoinetissier (changes requested):

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.ts
RaspberryChaos: Making this change is giving me `TypeError: Cannot assign to read only property 'mdx' of object`. Do you have a suggestion how I could fix this error @antoinetissier ? 

Capture d’écran 2022-12-02 113141

antoinetissier (approved):
Looks good ! ✅

PR #51, opened 2022-11-16
UI-7948 - Introduce aliases to allow using multiple versions of packages

RaspberryChaos:
When migrating from 4.3 to 5.0, we want the target types to be taken from @activeviam/activeui-sdk-5.0.x.
When migrating from 5.0 to 5.1, we want the input types to be taken from @activeviam/activeui-sdk-5.0.x and the output types to be taken from @activeviam/activeui-sdk-5.1.y, so we need to depend on different versions of the SDK.

This PR introduces version aliases to allow us to use different versions of the same package. It changes the names of the `@activeviam` 5.0 packages currently in the `dependencies` in `package.json` to aliases, for example `"@activeviam/activeui-sdk"` becomes `"@activeviam/activeui-sdk-5.0"`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,10 +30,10 @@
     "trailingComma": "all"
   },
   "dependencies": {
-    "@activeviam/activeui-sdk": "5.0.15",
-    "@activeviam/content-server-initialization": "5.0.15",
-    "@activeviam/data-model": "5.0.15",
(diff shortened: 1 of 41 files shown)


antoinetissier (approved):
Looks great !

PR #53, opened 2022-11-18
UI-7946 - CLI requires from-version and to-version arguments

eliasinho:
## Description

With this PR, the CLI now requires to new arguments: `from-version` (alias `f`), and `to-version`(alias `t`). The users hence now need to specify the versions they want to migrate their content from and to.

## Testing procedure

- Pull this branch and build the repo
- Outside of this repo, create a `migration-test` folder
- Init this folder (`npm init`) and install you local `activeui-migration` dependency using [`npm link`](https://www.geeksforgeeks.org/how-to-install-a-local-module-using-npm/)
- Follow the steps [here](https://github.com/activeviam/activeui-migration#installation) to finish the setup
- Run the migration command line adding the `-f` and `-t` arguments:
  - `-f 5.0 -t 5.1` should invalidate the `-f` argument
  - `-f 4.3 -t 5.1` should invalidate the `-t` argument
- if you want to test the case where both arguments are valid but don't reflect a valid combination (e.g. `-f 5.0 -t 4.3`), you would need to add other versions in the `migrationFunctions` array in `bin.ts`
- finally, the current default case `-f 4.3 -t 5.0` should still work


src/cli/bin.ts
```diff
@@ -95,19 +166,58 @@ yargs
...
+      if (fromVersionIndex >= toVersionIndex) {
+        throw new Error(`
+          You specified ${fromVersion} as the version to migrate your content from, and ${toVersion} as the version to migrate your content to.
+          The version to migrate to must be higher than the version to migrate from.
+        `);
+      }
+      const functionsToCompose = migrationFunctions
+        .slice(fromVersionIndex, toVersionIndex)
+        // The second element of the sub-arrays of the filtered `migrationFunctions` are systematically migration functions.
+        .map(([, func]) => func) as MigrationFunction[];
```
antoinetissier, line 207: You can avoid that by typing `migrationFunctions` as an array of tuples.

src/cli/bin.ts
```diff
@@ -95,19 +166,58 @@ yargs
...
+      toVersion,
       removeWidgets: keysOfWidgetPluginsToRemove,
       pivotInputPath,
       debug,
       stack,
     }) => {
+      const migrationFunctions = [["4.3", migrateUIFolder], ["5.0"]];
+      const versions = migrationFunctions.map(([version]) => version);
+      const fromVersions = versions.slice(0, -1);
+      const toVersions = versions.slice(1);
```
antoinetissier, line 179: This can be done outside of the function.

src/cli/bin.ts
```diff
@@ -95,19 +166,58 @@ yargs
...
       outputPath,
       serversPath,
+      fromVersion,
+      toVersion,
       removeWidgets: keysOfWidgetPluginsToRemove,
       pivotInputPath,
       debug,
       stack,
     }) => {
+      const migrationFunctions = [["4.3", migrateUIFolder], ["5.0"]];
```
antoinetissier, line 176: I'm not a fan of having the last element of the array being a different type: and the fact that it doesn't fit the `migrationFunctions` name reflects that.
Alternatives:
1. define a `latestMinorVersion` variable outside of the array
2. add `to` in the key, e.g. `"4.3_to_5.0"` instead of "4.3"

I like option 2. better, as it's understandable at a glance.

src/cli/bin.ts
```diff
@@ -95,19 +166,58 @@ yargs
...
+      const versions = migrationFunctions.map(([version]) => version);
+      const fromVersions = versions.slice(0, -1);
+      const toVersions = versions.slice(1);
+      if (!fromVersions.includes(fromVersion)) {
+        throw new Error(`
+          You specified ${fromVersion} as the version to migrate your content from. It's not a valid version to migrate from.
```
antoinetissier, line 182: Same but shorter:
```suggestion
          ${fromVersion} is not a valid version to migrate from.
```

src/cli/bin.ts
```diff
@@ -95,19 +166,58 @@ yargs
...
+          The valid versions to migrate from are: ${fromVersions.join(", ")}.
+        `);
+      }
+      if (!toVersions.includes(toVersion)) {
+        throw new Error(`
+          You specified ${toVersion} as the version to migrate your content to. It's not a valid version to migrate to.
```
antoinetissier, line 188:
```suggestion
          ${toVersion} is not a valid version to migrate to.
```

src/cli/bin.ts
```diff
@@ -95,19 +166,58 @@ yargs
...
+        .map(([, func]) => func) as MigrationFunction[];
+
       const legacyUIFolder = await fs.readJSON(inputPath);
       const legacyPivotFolder = pivotInputPath
         ? await fs.readJSON(pivotInputPath)
         : undefined;
       const servers = await fs.readJSON(serversPath);
 
-      const [migratedUIFolder, counters, errorReport] = await migrateUIFolder(
+      const [migratedUIFolder, counters, errorReport] = await compose(
```
antoinetissier, line 215: I'd find it clearer to add an intermediate step:
```suggestion
      const migrate = await compose(...functionsToCompose);
      const [migratedUIFolder, counters, errorReport] = await migrate(
```

src/cli/bin.ts
```diff
@@ -3,6 +3,63 @@ import _capitalize from "lodash/capitalize";
...
+    },
+  ]),
+  // _fromPairs returns a Dictionary.
+  // In this case, the keys used correspond to the attributes of OutcomeCounters.
+) as OutcomeCounters;
+
+const compose =
+  (...migrationFunctions: MigrationFunction[]) =>
+  async (
+    ...contentToMigrate: Parameters<MigrationFunction>
```
antoinetissier, line 49: I find this misleading cause it's not a `ContentRecord`.
It's okay to be abstract and use `...params` here.

src/cli/bin.ts
```diff
@@ -3,6 +3,63 @@ import _capitalize from "lodash/capitalize";
...
+    if (migrationFunctions.length === 0) {
+      return [contentToMigrate[0], initialCounters, {}];
+    }
+    const [migratedUIFolder, counters, errorReport] =
+      await migrationFunctions[0](...contentToMigrate);
+    return compose(...migrationFunctions.slice(1))(
+      migratedUIFolder,
+      counters,
+      errorReport || {},
+      contentToMigrate[3],
```
antoinetissier, line 60: This is hard to read.

antoinetissier (commented):
Instead of throwing errors inside the function called by the CLI, it would be nice to let the CLI perform the safety checks.
I'm sure there are examples online, and with a quick look https://yargs.js.org/docs/#choices and https://yargs.js.org/docs/#api-reference-checkfn-globaltrue seem like good candidates.

src/cli/bin.ts
```diff
@@ -3,6 +3,63 @@ import _capitalize from "lodash/capitalize";
...
+  (...migrationFunctions: MigrationFunction[]) =>
+  async (
+    ...contentToMigrate: Parameters<MigrationFunction>
+  ): ReturnType<MigrationFunction> => {
+    if (migrationFunctions.length === 0) {
+      return [contentToMigrate[0], initialCounters, {}];
+    }
+    const [migratedUIFolder, counters, errorReport] =
+      await migrationFunctions[0](...contentToMigrate);
+    return compose(...migrationFunctions.slice(1))(
```
antoinetissier, line 56: No need to slice, you can destructure instead which is more efficient: `[firstFunction, ...rest]`.

But it seems to me that an implementation with `reduce` or even a `for` loop would be shorter ?

eliasinho:
> If possible, instead of throwing errors inside the function called by the CLI, it would be nice to let the CLI perform the safety checks. I'm sure there are examples online, and with a quick look https://yargs.js.org/docs/#choices and https://yargs.js.org/docs/#api-reference-checkfn-globaltrue seem like good candidates.

Good idea! `choices` indeed validates that the arguments belong to the right set of versions.

However to validate that `from-version` is older than `to-version`, their `check` function does enables to actually check against the arguments, and implement our logic, but I don't see a big added value in having the logic in `check` rather than in the core `argument` callback. I think we want to throw with a clear message of ours, so it's just a copy-paste of what was done before, and it makes us redefine `fromVersionIndex` and `toVersionIndex`.

I don't feel super strongly about it, so I made the change, but yeah, I don't see any improvement using `check`. Let me know what you think!

src/cli/bin.ts
eliasinho: Related to what's coming up in [this PR](https://github.com/activeviam/activeui-migration/pull/55), I changed the signature of `migrate_43_to_50` and used destructuration, which makes the `compose` simpler. Let me know what you think.

src/cli/bin.ts
eliasinho: I'm not a fan of 2. in the current state, because that means that we would have to parse the `"fromVersion_to_toVersion"` string to extract `fromVersions` and `toVersions`.
Since your comment was written, `migrateUIFolder` was renamed into `migrate_43_to_50`, and `migrationFunctions` typed as a tuple, so what the array represents is easily understandable IMO.

However I agree with not having the last element of the array being of a different type. I see two possibilities:
3. the last migration function is a no-op function 
4. add `to` to the key as you suggested in 2., but have another hardcoded `versions` array on the side to avoid the parsing of the string

I prefer 3., to avoid having two different arrays and have only one source of truth.
What do you think?

I haven't made any change for now

src/cli/bin.ts
```diff
@@ -175,6 +286,21 @@ This will output a file named \`report.json\` containing the error messages.`);
...
+      ([version]) => version === fromVersion,
+    );
+    const toVersionIndex = migrationFunctions.findIndex(
+      ([version]) => version === toVersion,
+    );
+    if (fromVersionIndex >= toVersionIndex) {
+      throw new Error(`
+        You specified ${fromVersion} as the version to migrate your content from, and ${toVersion} as the version to migrate your content to.
+        The version to migrate to must be higher than the version to migrate from.
+      `);
```
antoinetissier, line 300: I would keep it simple:
```suggestion
      throw new Error("--to-version must be greater than --from-version");
```

src/cli/bin.ts
```diff
@@ -175,6 +286,21 @@ This will output a file named \`report.json\` containing the error messages.`);
...
     },
   )
+  .check(({ fromVersion, toVersion }) => {
+    const fromVersionIndex = migrationFunctions.findIndex(
+      ([version]) => version === fromVersion,
+    );
+    const toVersionIndex = migrationFunctions.findIndex(
+      ([version]) => version === toVersion,
+    );
+    if (fromVersionIndex >= toVersionIndex) {
```
antoinetissier, line 296: If possible let's just use `semver`:
```suggestion
    if (semver.gte(fromVersion, toVersion)) {
```

src/cli/bin.ts
```diff
@@ -1,8 +1,71 @@
...
+
+const migrationFunctions: [
+  versionFrom: string,
+  migrationFunction?: MigrationFunction,
+][] = [["4.3", migrate_43_to_50], ["5.0"]];
+const versions = migrationFunctions.map(([version]) => version);
+const fromVersions = versions.slice(0, -1);
+const toVersions = versions.slice(1);
+
+const compose =
```
antoinetissier, line 46: I think the recursive implementation adds unnecessary complexity here.
I would go for `reduce` or a `for` loop, and actually we could inline it instead of going for a `compose` abstraction since it'll be used in only one place.

src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -211,33 +208,23 @@ const accumulateStructure = ({
 export async function migrate_43_to_50(
   legacyUIFolder: ContentRecord,
   {
+    migratedUIFolder,
```
antoinetissier, line 211: It seems wrong to have both `legacyUIFolder` and `migratedUIFolder`.
I thought we had agreed on passing `contentServer` down the line, and that each function would mutate it.
If so then here we shouldn't have en extra `migratedUIFolder` argument, and `legacyUIFolder` should be renamed to `uiFolder`

src/cli/bin.ts
```diff
@@ -1,8 +1,71 @@
...
+      [serverKey: string]: {
+        dataModel: DataModel;
+        url: string;
+      };
+    };
+    keysOfWidgetPluginsToRemove?: string[] | undefined;
+    doesReportIncludeStacks: boolean;
+    legacyPivotFolder?: ContentRecord<any> | undefined;
+  },
+) => Promise<void>;
```
antoinetissier, line 36: Related to my other comment on `migrate_43_to_50`: you intend to do the refactoring that we talked about in another PR right ?
Because for example we shouldn't pass just the UI folder but the full content server.

antoinetissier (commented):
There shouldn't be changes to the `bin.js` file, and I think back in the day we had issues because of that, so let's revert them

src/cli/bin.ts
eliasinho: Yes, I'll do it in [this PR](https://github.com/activeviam/activeui-migration/pull/55), if that's fine by you.

src/cli/bin.ts
eliasinho: Ah yes, I like the structure in 
```
{
  from: string;
  to: string;
  migrate: MigrationFunction;
}[]
```
Not sure about having each individual object exported from the corresponding migration function, I kind of like having a centralized array with all the versions and corresponding migration functions explicitly written.

eliasinho:
> There shouldn't be changes to the `bin.js` file, and I think back in the day we had issues because of that, so let's revert them

I think I'l need to do a force push for that one, do you mind?

src/cli/bin.ts
```diff
@@ -69,6 +102,20 @@ yargs
         demandOption: true,
         desc: "The path to the JSON file holding the servers information.",
       });
+      args.option("from-version", {
+        alias: "f",
+        type: "string",
+        demandOption: true,
+        choices: fromVersions,
+        desc: "The version which the content is migrated from.",
```
antoinetissier, line 110: We can be even more concise I think:
```suggestion
        desc: "The version to migrate from.",
```

src/cli/bin.ts
```diff
@@ -69,6 +102,20 @@ yargs
...
+      });
+      args.option("to-version", {
+        alias: "t",
+        type: "string",
+        demandOption: true,
+        choices: toVersions,
+        desc: "The version which the content is migrated to.",
```
antoinetissier, line 117:
```suggestion
        desc: "The version to migrate to.",
```

src/cli/bin.ts
```diff
@@ -182,6 +241,12 @@ This will output a file named \`report.json\` containing the error messages.`);
       process.exit(0);
     },
   )
+  .check(({ fromVersion, toVersion }) => {
+    if (gte(toVersion, fromVersion)) {
```
antoinetissier, line 245:
```suggestion
    if (lte(toVersion, fromVersion)) {
```

src/cli/bin.ts
```diff
@@ -114,20 +162,31 @@ yargs
...
+      const fromVersionIndex = migrationFunctions.findIndex(
+        ({ from }) => from === fromVersion,
+      );
+      const toVersionIndex = migrationFunctions.findIndex(
+        ({ to }) => to === toVersion,
+      );
+
+      for (const { migrate } of migrationFunctions.slice(
+        fromVersionIndex,
+        toVersionIndex,
```
antoinetissier, line 176:
```suggestion
        toVersionIndex + 1,
```

antoinetissier (commented):
Without the changes that I propose below I would expect the CLI not to work.
Have you tested it ?

antoinetissier:
> I think I'll need to do a force push for that one, do you mind?

Go ahead 🟢

eliasinho:
> > I think I'll need to do a force push for that one, do you mind?
> 
> Go ahead 🟢

Actually, didn't a force push, all good 👌

src/cli/bin.ts
```diff
@@ -182,6 +241,12 @@ This will output a file named \`report.json\` containing the error messages.`);
       process.exit(0);
     },
   )
+  .check(({ fromVersion, toVersion }) => {
+    if (gte(fromVersion, toVersion)) {
```
eliasinho, line 245: Actually @antoinetissier this doesn't work, `semver` throws an `Invalid Version` error. I did some testing with it and it seems that when passing `X.Y` as the version, we're missing the patch version for the format to be correct.

It raises the question of what format we accept for the versions. I think we should not be too flexible and only accept `X.Y` formats. Then we can manually append an arbitrary `0` patch version to the versions before `semver` compares them.

What do you think?

antoinetissier: You can use `semver.coerce`

eliasinho: Thanks!

eliasinho:
@antoinetissier I just tested the different cases and all seems good!

antoinetissier (approved):
Looks good ✅

RaspberryChaos (approved):
Looks good to me ! It worked as expected when tested👍

PR #55, opened 2022-11-23
UI-8047 - Introduce generic getMigrateDashboards function

eliasinho:
## Description

This PR introduces a new `getMigrateDashboards` function, supposed to be generic, that: 
- takes a `cb` callback argument and calls it. This callback function handles the actual migration of the dashboard, the logic in it depending on which migration is done
- updates `counters` and `migratedUIFolder`
- handles error cases by mutating `errorReport`
 
In `bin.ts`, a curried `migrateDashboards` function will be created thanks to `getMigrateDashboards` and passed to `migrate_fromVersion_to_toVersion`. Global variables meant to be mutated are also defined at the level of `bin.ts` and are passed to `migrate_fromVersion_to_toVersion`.

In `migrate_fromVersion_to_toVersion`, the dashboards migration is handled in a callback passed to `migrateDashboards`, passed in the arguments. `migrate_fromVersion_to_toVersion` returns a `void`.

The compose of different migration fonctions is handled in [another PR](https://github.com/activeviam/activeui-migration/pull/53).


src/getMigrateDashboards.ts
```diff
@@ -0,0 +1,86 @@
...
+    doesReportIncludeStacks: boolean;
+  },
+  cb: MigrateDashboardCallback<fromVersion>,
+): void => {
+  let migratedDashboard;
+
+  const legacyContent =
+    legacyUIFolder.children?.dashboards.children?.content.children;
+  const legacyStructure =
+    legacyUIFolder?.children?.bookmarks?.children?.structure!;
```
antoinetissier, line 46: I don't understand the changes here.
- If this is intended to work for 4.3, this definition of `legacyContent` should not work: there is no `ui/dashboards` folder in 4.3.
- If this is intended to work for 5.0, then this definition of `legacyStructure` shold not work: there is no `ui/bookmarks` in 5.0.
- It doesn't make sense to mix both 4.3 and 5.0.

We discussed that 4.3 to 5.0 is a special case that cannot really be genericized anyway, and so that:
- `getMigrateDashboards` would assume the ActiveUI 5 content server structure, and be used to migrate for 5.x to 5.y
- 4.3 to 5.0 would handle this logic on its own, by mutating `contentServer` directly.

Am I missing something ?

src/getMigrateDashboards.ts
```diff
@@ -0,0 +1,86 @@
...
+  MigrateFromVersions,
+  OutcomeCounters,
+} from "./migration.types";
+import { _addErrorToReport } from "./_addErrorToReport";
+import { _getMapOfFolderIds } from "./_getMapOfFolderIds";
+import { _serializeError } from "./_serializeError";
+
+type MigrateDashboardCallback<fromVersion extends MigrateFromVersions> = (
+  dashboardState: fromVersion extends "5.0" ? DashboardState50 : any,
+) => fromVersion extends "5.0" ? DashboardState51 : any;
```
antoinetissier, line 18: It's equivalent but I would find it clearer this way:
```
) => toVersion extends "5.1" ? DashboardState51 : any;
```

But anyway even though I understand the intent I don't think we should centralize the types here.
Because:
- I think it's a typical case where it's better to colocate the types necessary for each step in the step itself
- In any case, I expect these types to be lost and to require TypeScript workarounds when composing higher up

Rather, we would type things correctly when implementing `migrateDashboard` in each version.
E.g. in `5.0` to `5.1`:
```
function migrateDashboard(legacyDashboardState: DashboardState50): DashboardState51 {...}
```

src/getMigrateDashboards.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Handles the generic logic to migrate dashboards, specifically the errors.
+ * `cb` is a generic callback that is meant to handle dashboards migration from one specific version to another.
+ * Mutates `migratedUIFolder`, `dashboards`, `counters` and `errorReport`.
+ */
```
antoinetissier, line 24: If we go for: "`getMigrateDashboards` returns a function", then I would suggest:
```suggestion
/**
 * Returns a function which can be called to migrate the content of each dashboard, according to the logic defined in its `callback` argument.
 *
 * The logic of traversing each file in `/ui/dashboards` and the error handling is encapsulated to make it easier for the caller.
 *
 * Mutates `contentServer`, `counters` and `errorReport`.
 */
```

src/getMigrateDashboards.ts
```diff
@@ -0,0 +1,86 @@
...
+      migratedDashboard = dashboard;
+    }
+
+    migratedUIFolder.children!.dashboards.children!.content.children![fileId] =
+      {
+        entry: {
+          ...entry,
+          content: JSON.stringify(_omit(migratedDashboard, ["name"])),
+        },
+      };
```
antoinetissier, line 84: This logic should depend on the `behavior` flag down the line.
Do you intend to add it in a separate PR ?

src/cli/bin.ts
```diff
@@ -100,21 +105,38 @@ yargs
       debug,
       stack,
     }) => {
+      const emptyUIFolder = emptyUIFolder50;
+      const migratedUIFolder: ContentRecord = _cloneDeep(emptyUIFolder);
```
antoinetissier, line 109: We shouldn't have this.
Each migration function should just receive `contentServer`.

src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -211,33 +204,23 @@ const accumulateStructure = ({
...
+    migratedUIFolder: ContentRecord;
+    errorReport: ErrorReport;
+    counters: OutcomeCounters;
     servers: { [serverKey: string]: { dataModel: DataModel; url: string } };
     keysOfWidgetPluginsToRemove?: string[];
     doesReportIncludeStacks: boolean;
     legacyPivotFolder?: ContentRecord;
   },
-): Promise<[ContentRecord, OutcomeCounters, ErrorReport?]> {
-  const migratedUIFolder: ContentRecord = _cloneDeep(emptyUIFolder);
```
antoinetissier, line 225: This logic should not be moved outside for `4.3` to `5.0`.

src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -211,33 +204,23 @@ const accumulateStructure = ({
 export async function migrate_43_to_50(
   legacyUIFolder: ContentRecord,
   {
+    migratedUIFolder,
```
antoinetissier, line 207: This shouldn't be needed.
Like every migration function, it should just mutate an input `contentServer`.

antoinetissier (commented):
Unless I'm missing something the logic of `_getFolderName` is specific to `4.3` and it should stay under `4.3_to_5.0`.

src/getMigrateDashboards.ts
eliasinho: Yes, I was intending to handle the `behavior` part in a separate PR. Here I'm implementing kind of a "default" behavior.

src/getMigrateDashboards.ts
eliasinho: Okay, so you would see the callback type as a union of all migration callbacks?

src/getMigrateDashboards.ts
```diff
@@ -0,0 +1,86 @@
...
+type MigrateDashboardCallback<fromVersion extends MigrateFromVersions> = (
+  dashboardState: fromVersion extends "5.0" ? DashboardState50 : any,
+) => fromVersion extends "5.0" ? DashboardState51 : any;
+
+/**
+ * Handles the generic logic to migrate dashboards, specifically the errors.
+ * `cb` is a generic callback that is meant to handle dashboards migration from one specific version to another.
+ * Mutates `migratedUIFolder`, `dashboards`, `counters` and `errorReport`.
+ */
+export const getMigrateDashboards = <fromVersion extends MigrateFromVersions>(
```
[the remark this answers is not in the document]
eliasinho: Agreed on the naming being confusing at the moment, I'll change that.
Regarding where the currying is done, I don't see a huge difference between 1 and 2, but I'm happy to go for 1.

src/getMigrateDashboards.ts
eliasinho: Yes, you're completely right, that's my bad! I fixed that

src/migration.types.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Report of errors that occurred during the migration of the /ui folder.
+ * Output in `./report.json` if the user activates the `--debug` option.
+ */
+export interface ErrorReport {
+  dashboards?: {
+    [
+      dashboardId: string
+    ]: // If the error was thrown by `migrateDashboard` itself, not an underlying call to `migrateWidget`.
+    // This should happen rarely.
+    // In this case, the whole outdated dashboard is copied as is.
```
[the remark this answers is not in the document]
eliasinho: Yes, and yes!

src/getMigrateDashboards.ts
antoinetissier: We could type it this way:
```
type MigrateDashboardCallback<FromDashboardState, ToDashboardState> = (dashboardState: FromDashboardState) => ToDashboardState;
```

It's just that this doesn't add a lot of value since it's implicitly `any` for both generics.

What matters is that each `migrateDashboardFunction` under each folder is typed properly, like:
```
function migrateDashboard(legacyDashboardState: DashboardState50): DashboardState51 {...}
```

src/getMigrateDashboards.ts
eliasinho: By the way, now that I think of it, it doesn't mutate those variables, the mutation is done by the migration function that is returned. I'll update the comment

src/getMigrateDashboards.ts
```diff
@@ -0,0 +1,81 @@
...
+          name: dashboard.name,
+        });
+
+        migratedDashboard = dashboard;
+      }
+
+      dashboardsContent![fileId] = {
+        entry: {
+          ...entry,
+          content: JSON.stringify(_omit(migratedDashboard, ["name"])),
```
antoinetissier, line 77: Omitting the name is specific to 4.3 => 5.0 as well.

src/migration.types.ts
```diff
@@ -0,0 +1,87 @@
...
+  filters?: {
+    [filterId: string]: FileErrorReport;
+  };
+  folders?: {
+    [folderId: string]: FileErrorReport;
+  };
+}
+
+type MigrateDashboards50To51 = (
+  dashbaordState: DashboardState50,
```
RaspberryChaos, line 81:
```suggestion
  dashboardState: DashboardState50,
```

src/getMigrateDashboards.ts
```diff
@@ -41,7 +42,7 @@ export const getMigrateDashboards =
     const dashboardsStructure =
       contentServer.children?.ui.children?.dashboards.children?.structure!;
 
-    const mapOfFolderIds = _getMapOfFolderIds(dashboardsStructure);
+    const mapOfFolderIds = _getFilesAncestry(dashboardsStructure);
```
antoinetissier, line 45: The name is not appropriate anymore: it should be `filesAncestry` or `mapOfAncestors`

src/_getFilesAncestry.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+}
+
+/**
+ * Iterates over `structure` and returns a map from file id to the ids and names of the parent folders of the file.
+ */
+export function _getFilesAncestry(structure: ContentRecord): {
+  [fileId: string]: { id: string; name: string }[];
+} {
+  const mapOfFolderIds: { [id: string]: { id: string; name: string }[] } = {};
```
antoinetissier, line 27: The name is not appropriate anymore.

src/_getFilesAncestry.ts
```diff
@@ -0,0 +1,32 @@
+import { ContentRecord } from "@activeviam/activeui-sdk-5.0";
+
+function addChildren(
+  node: ContentRecord,
+  path: { id: string; name: string }[],
+  contentRecordToPath: { [id: string]: { id: string; name: string }[] },
```
antoinetissier, line 6: This is mutated during the recurstion, and it's more `fileIdToPath` than `contentRecordToPath`.
So I would call it `ancestryAccumulator` or just `accumulator` since it's used only internally in this file for the recursion.

src/_getFilesAncestry.ts
```diff
@@ -0,0 +1,32 @@
+import { ContentRecord } from "@activeviam/activeui-sdk-5.0";
+
+function addChildren(
```
antoinetissier, line 3: I find `addChildren` misleading, as we're not adding the child nodes themselves:
```suggestion
function accumulateAncestryOfChildren(
```

src/_getFilesAncestry.ts
```diff
@@ -0,0 +1,32 @@
...
+        addChildren(child, [...path, { id, name }], contentRecordToPath);
+      }
+    });
+  }
+}
+
+/**
+ * Iterates over `structure` and returns a map from file id to the ids and names of the parent folders of the file.
+ */
+export function _getFilesAncestry(structure: ContentRecord): {
```
antoinetissier, line 24: Let's add a test for this function.

src/getMigrateDashboards.ts
```diff
@@ -32,7 +31,9 @@ export const getMigrateDashboards =
       doesReportIncludeStacks: boolean;
     },
   ) =>
-  (callback: MigrateDashboardsCallback): void => {
+  <FromDashboardState, ToDashboardState>(
+    callback: MigrateDashboardsCallback<FromDashboardState, ToDashboardState>,
```
antoinetissier, line 35: It is indeed a callback passed to the curried `migrateDashboards` function.
But its contract is to migrate a single dashboard, so I would call it `MigrateDashboardCallback`.
What do you think ?

src/_getFilesAncestry.ts
eliasinho: Fair point, I'm happy with `accumulator`

src/_getFilesAncestry.ts
eliasinho: Good point, although I prefer `accumulateChildrenAncestors`, I find it easier to read, what do you think?

src/getMigrateDashboards.ts
eliasinho: Yeah, I agree. The callback migrates only one dashboard, so it's type should reflect that, good point 👍

src/__test_resources__/dashboardsStructure.ts
```diff
@@ -0,0 +1,1994 @@
+import { ContentRecord } from "@activeviam/activeui-sdk-5.0";
+
+/**
+ * Example of the dashboard's structure part of a content server's ui folder.
+ * Can be found on the content server under /ui/dashboards/structure. 
+ */
+export const dashboardsStructure: ContentRecord = {
```
eliasinho, line 7: I divided the length of this file by 2, but it can be shortened again if needed.

src/_getFilesAncestry.ts
antoinetissier: I think a "child ancestor" could be confusing. But what about `accumulateAncestorsOfChildren` or just `accumulateAncestors` then ?
If we go for that, then we could align on it with `_getFileAncestors`

eliasinho: As discussed, let's go for `accumulateAncestry` and keep `_getFilesAncestry`

antoinetissier (commented):
Looks good, just some remarks on the test resources.

It would be nice to keep it very small initially: < 5 items for each type of saved content (widgets, dashboards, filters, calculated measures), ideally each time with at least one item in a folder.
If we do this I'd expect the order of magnitude of the snapshot size to go from several thousands to several hundreds.

I also think we don't need both `contentServer` and `smallContentServer`: at least for now I'd keep just `smallContentServer` (and we could call it `contentServer`).
A big one like we have on the ranch could maybe be useful to test the CLI, but even then I don't think anyone would check a 10 000 lines snapshot.

Finally, I think these test resources should go under 5.0_to_5.1 since it's a 5.0 content server.
To be clear, `_getFilesAncestry.test.ts` would stay out of 5.0_to_5.1, but `smallContentServer` would be in it.

If you prefer to handle all of this outside of this PR, you can just inline a small `dashboardsStructure` in `_getFileAncestry.test.ts`, up to you !

RaspberryChaos:
I added a `uiWidgetsFolder` and a `uiDashboardsFolder` to the `5.0_to_5.1` test resources to test migrating calculated measures in widgets and dashboards. Inside `contentServer` we could potentially use:

```
        widgets: uiWidgetsFolder,
        dashboards: uiDashboardsFolder,
```

This would make it smaller, and they both have at least one item in a folder as suggested by @antoinetissier. What do you think, @eliasinho ?

eliasinho:
> I added a `uiWidgetsFolder` and a `uiDashboardsFolder` to the `5.0_to_5.1` test resources to test migrating calculated measures in widgets and dashboards. Inside `contentServer` we could potentially use:
> 
> ```
>         widgets: uiWidgetsFolder,
>         dashboards: uiDashboardsFolder,
> ```
> 
> This would make it smaller, and they both have at least one item in a folder as suggested by @antoinetissier. What do you think, @eliasinho ?

Ah yes, definitely, thanks @RaspberryChaos!

RaspberryChaos:
I also created a `uiCalculatedMeasuresFolder` test resource, which I had added to another PR but ended up deleting [here](https://github.com/activeviam/activeui-migration/pull/45/commits/28ab2e711e005220d0200079084105f10167734e). I was planning to add it for testing #59, but you could also copy and paste and add it here if you think it would be useful.

eliasinho:
Actually the test resources already existing under `5.0_to_5.1/__test_resources__` are a bit bigger than what we would like (less than 5 items per saved content type). So I just went for reducing the size of `smallContentServer`, which is 1000+ lines now. It does contain less than 5 items per content types.
Let me know if you think it looks good @antoinetissier and @RaspberryChaos!

Also, I had a question, since it's a `5.0_to_5.1` test resource, we don't want anything under `/pivot/cm`, do we?

antoinetissier:
> Also, I had a question, since it's a 5.0_to_5.1 test resource, we don't want anything under /pivot/cm, do we?

It would be nice to have at least one calculated measure in there, to check that the ones already defined in the cube are preserved when we migrate the legacy CMs from the /ui folder to /pivot.

You can do it here if you want, but @RaspberryChaos can also do it when adding the final PR and test for `migratCalculatedMeasures`. Up to you guys !

eliasinho:
> > Also, I had a question, since it's a 5.0_to_5.1 test resource, we don't want anything under /pivot/cm, do we?
> 
> It would be nice to have at least one calculated measure in there, to check that the ones already defined in the cube are preserved when we migrate the legacy CMs from the /ui folder to /pivot.
> 
> You can do it here if you want, but @RaspberryChaos can also do it when adding the final PR and test for `migratCalculatedMeasures`. Up to you guys !

Actually I already have a couple calculated measures there, I just wanted to check if we wanted to keep them or not!

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
```diff
@@ -0,0 +1,1240 @@
...
+                        readers: ["admin"],
+                        timestamp: 1669805699247,
+                        lastEditor: "admin",
+                        canRead: true,
+                        canWrite: true,
+                      },
+                    },
+                  },
+                },
+                aca: {
```
antoinetissier, line 303: I see `aca` and `9a7` in `structure` but not in `content`, even though as their metadata file suggests they're not folders.
Let's adapt this.
And actually, it would be nice to have a calculated measure under a folder.

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
```diff
@@ -0,0 +1,1240 @@
...
+                                isDirectory: false,
+                                owners: ["admin"],
+                                readers: ["admin"],
+                                timestamp: 1669817918087,
+                                lastEditor: "admin",
+                                canRead: true,
+                                canWrite: true,
+                              },
+                            },
+                          },
```
antoinetissier, line 1078: It seems to me that `c9c` is in `structure` but neither in `content` nor `thumbnails`.
If so, can you remove it ?

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
```diff
@@ -0,0 +1,1240 @@
...
+                                readers: ["admin"],
+                                timestamp: 1669817950113,
+                                lastEditor: "admin",
+                                canRead: true,
+                                canWrite: true,
+                              },
+                            },
+                          },
+                        },
+                        "478": {
```
antoinetissier, line 1018: From a quick look it looked like `478` is a child of `346` in `structure`, but of `31b` in `thumbnails`.
Can you double check and adapt if necessary ?

antoinetissier (commented):
Almost there ! Just some last questions on the test resource.

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
eliasinho: Nope, it's a child of `31b` in `structure` too:
image


image

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
eliasinho: Yes, good catch, my bad!

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
eliasinho: Should be good now!

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
```diff
@@ -0,0 +1,1240 @@
...
+        readers: ["ROLE_CS_ROOT"],
+        timestamp: 1669226473908,
+        lastEditor: "_no_user_",
+        canRead: true,
+        canWrite: true,
+      },
+      children: {
+        __default__: {
+          entry: {
+            content: '{"owners":["ROLE_ADMIN"],"readers":["__ALL_USERS__"]}',
```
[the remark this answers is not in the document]
eliasinho: I originally just exported the whole content server from Admin UI, so that's where it comes from. Not sure why it's `"__ALL_USERS__"` and not `"ROLE_USER"`.

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
```diff
@@ -0,0 +1,1240 @@
+import { ContentRecord } from "@activeviam/activeui-sdk-5.0";
+
+/**
+ * Content server object, useful for tests.
+ * Lighter version of `contentServer` in `./contentServer.ts`.
+ */
+export const smallContentServer: ContentRecord = {
```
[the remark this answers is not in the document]
eliasinho: No actually you're right, let's rename it to `contentServer`.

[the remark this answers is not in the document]
eliasinho: However, the risk I see is that with time, people enrich it with new content, to fit their need for new test use cases. And it might get bigger and not stay "small". Hence making clear now that this object is supposed to be the "small" content server might better.
Although as it concerns only the 5.0 migration, it might no be that much of a worry.
What do you think @antoinetissier?

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
antoinetissier: Thanks for the double check

src/5.0_to_5.1/__test_resources__/smallContentServer.ts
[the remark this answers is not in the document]
antoinetissier: > However, the risk I see is that with time, people enrich it with new content, to fit their need for new test use cases.

If that's incremental like this then I think it's justified and not too much of an issue: you're going to add one specific dashboard to fit your test use case for example.

Even in the event where for some reason someone would try to merge a big structure on top of the `smallContentServer`, it should be caught by the review imo.

So personally I have a preference for `contentServer` for now, and only adapting later if needed.

antoinetissier (approved):
Looks great, let's go !

PR #56, opened 2022-12-01
UI-8014 - migrate_43_to_50 only accepts and mutates contentServer, counters and errorReport

eliasinho:
## Description

At the moment, `migrate_43_to_50` accepts `legacyUIFolder` as an argument, and returns the newly `migratedUIFolder`.
It also takes `legacyPivotFolder` as an argument.
It also creates and returns the `counters` and the `errorReport`.

For migration functions to be generically composable, a simpler design would be to have it:
* take the whole content server as an argument: no need for `legacyUIFolder` and `legacyPivotFolder` anymore
* mute the content server object inside the function
* accept `counters` and `errorReport` as arguments and mutate them inside
* return a `Promise<void>`

### DIFF
--- a/src/4.3_to_5.0/migrate_43_to_50.ts
+++ b/src/4.3_to_5.0/migrate_43_to_50.ts
@@ -200,7 +200,7 @@ const accumulateStructure = ({
 };
 
 /**
- * Modifies the ActiveUI 4.3 `contentServer` structure, for it to be ready to be used by ActiveUI 5.0.
+ * Migrates `contentServer` from a version usable by ActiveUI 4.3 to one usable by ActiveUI 5.0.
  * Also keeps the number of migration successes and failures in `counters` and a detailed `errorReport`.
  *
  * Widgets with keys in `keysOfWidgetPluginsToRemove` are not migrated:
--- a/src/4.3_to_5.0/migrate_43_to_50.ts
+++ b/src/4.3_to_5.0/migrate_43_to_50.ts
@@ -227,7 +227,7 @@ export async function migrate_43_to_50(
 ): Promise<void> {
   if (contentServer.children?.ui === undefined) {
     throw new Error(
-      "Your content server structure doesn't contain any ui folder.",
+      "Your content server doesn't contain any /ui folder.",
     );
   }
 
--- a/src/4.3_to_5.0/migrate_43_to_50.ts
+++ b/src/4.3_to_5.0/migrate_43_to_50.ts
@@ -201,7 +201,7 @@ const accumulateStructure = ({
 
 /**
  * Migrates `contentServer` from a version usable by ActiveUI 4.3 to one usable by ActiveUI 5.0.
- * Also keeps the number of migration successes and failures in `counters` and a detailed `errorReport`.
+ * Also keeps track of the number of migration successes and failures in `counters` and a detailed `errorReport`.
  *
  * Widgets with keys in `keysOfWidgetPluginsToRemove` are not migrated:
  * - for a matching saved ActiveUI 4.3 widget, no ActiveUI 5.0 file is created.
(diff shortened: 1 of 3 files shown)


src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -201,42 +200,39 @@ const accumulateStructure = ({
...
+ * - for a saved ActiveUI 4.3 dashboard including a matching widget, the widget is removed from the output ActiveUI 5.0 dashboard, and the layout is adapted so that siblings take the remaining space.
+ *
+ * Mutates `contentServer`, `errorReport` and `counters`.
  */
 export async function migrate_43_to_50(
-  legacyUIFolder: ContentRecord,
+  contentServer: ContentRecord,
   {
+    errorReport,
+    counters,
```
mtadams007, line 216: What's the logic in adding these as arguments instead of creating them from scratch?

mtadams007 (commented):
Had a question.

src/4.3_to_5.0/migrate_43_to_50.ts
eliasinho: They now represent global variables that might be used throughout successive migrations. For example, consider a 4.3 --> 5.1 migration. We'll run `migrate_43_to_50`, then (the not yet existing) `migrate_50_to_51`. In the end, we want `counters` and `errorReport` to contain information from both migrations.

Hence why we want to define them at a higher level, and populate them (by mutating them) inside, between other migration functions, `migrate_43_to_50`.

src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -201,42 +200,39 @@ const accumulateStructure = ({
 };
 
 /**
- * Returns the converted UI folder, ready to be used by ActiveUI 5.
- * Also returns a count of the number of migration successes and failures and an detailed error report.
+ * Modifies the ActiveUI 4.3 `contentServer` structure, for it to be ready to be used by ActiveUI 5.0.
+ * Also keeps the number of migration successes and failures in `counters` and a detailed `errorReport`.
```
antoinetissier, line 204: I'll let @mtadams007 confirm but "keeps" alone seems wrong here:
```suggestion
 * Also keeps track of the number of migration successes and failures in `counters` and a detailed `errorReport`.
```

src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -201,42 +200,39 @@ const accumulateStructure = ({
...
     servers: { [serverKey: string]: { dataModel: DataModel; url: string } };
     keysOfWidgetPluginsToRemove?: string[];
     doesReportIncludeStacks: boolean;
-    legacyPivotFolder?: ContentRecord;
   },
-): Promise<[ContentRecord, OutcomeCounters, ErrorReport?]> {
+): Promise<void> {
+  if (contentServer.children?.ui === undefined) {
+    throw new Error(
+      "Your content server structure doesn't contain any ui folder.",
```
antoinetissier, line 230:
```suggestion
      "Your content server doesn't contain any /ui folder.",
```

src/4.3_to_5.0/migrate_43_to_50.ts
eliasinho: Ah yes, my bad!

mtadams007: Yep @antoinetissier's wording is good

antoinetissier (approved):
Looks good ✅ 
We'll have to mention the change of CLI in the changelog, but it's probably more convenient to do it at the end once we have merged all PRs related to it.

mtadams007 (approved):
:rocket:

PR #58, opened 2022-12-12
UI-8048 - Introduce getMigrateWidgets function

eliasinho:
## Description

Introduce a `getMigrateWidgets` function, supposed to be generic, that:

- returns a migration function that takes a `callback` argument. This callback function handles the actual migration of the widgets, the logic in it depending on which migration is done
- mutates `contentServer` and `counters`.
- handles error cases by mutating `errorReport`

In `bin.ts`, a curried `migrateWidgets` function will be created thanks to `getMigrateWidgets ` and passed to `migrate_fromVersion_to_toVersion`. The global variables meant to be mutated are also defined at the level of `bin.ts` and are passed to `migrate_fromVersion_to_toVersion`.

In `migrate_fromVersion_to_toVersion`, the widgets migration is handled in the callback passed to `migrateWidgets`, passed in the arguments. `migrate_fromVersion_to_toVersion` returns a `void`.


antoinetissier:
See the review on https://github.com/activeviam/activeui-migration/pull/58

eliasinho:
> Additionally, we need the logic of migrating each widget in each dashboard.

Not sure what you mean here?

antoinetissier (approved):
The first iteration looks good to me, we'll add more things in subsequent PRs ✅

PR #59, opened 2022-12-13
UI-7834 - Migrate calculated measures from content server to ActivePivot

RaspberryChaos:

In AUI 5.0 calculated measures are saved to the content server.
In 5.1 we want to publish them directly to the cube.
Users migrating to 5.1 who have saved measures need to be able to migrate them to the cube.

This PR adds a function to migrate the calculated measures saved in ActiveUI 5.0. It also removes the definitions of calculated measures from saved dashboards and saved widgets.


src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.ts
```diff
@@ -0,0 +1,123 @@
...
+        JSON.parse(record.entry.content),
+        measureName,
+      );
+      record.entry.content = JSON.stringify(migratedContent);
+
+      cubeNames.forEach((cubeName) => {
+        if (cmFolder && cmFolder.children) {
+          // If `cubeName` property does not already exist, create it.
+          if (!cmFolder.children[cubeName]) {
+            cmFolder.children[cubeName] = {
```
RaspberryChaos, line 99: Will the `cubeName` already exist in the `pivot/entitlements/cm` folder or will there be instances where we need to create it ? I can delete this if it is not necessary.

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.ts
```diff
@@ -0,0 +1,123 @@
...
+  } = migrateCalculatedMeasuresInDashboards(
+    contentServer.children?.ui.children?.dashboards!,
+    dataModel,
+    namesOfCalculatedMeasuresToMigrate,
+  );
+
+  const measureToCubeMapping = _merge(
+    measureToCubeMappingInWidgets,
+    measureToCubeMappingInDashboards,
+  );
```
antoinetissier, line 79: `_merge` will not work:
image

Instead, we want to do the union of the arrays for each entry:
```suggestion
  const measureToCubeMapping = namesOfCalculatedMeasuresToMigrate.reduce((acc, measureName) => {
  const cubes = _uniq([...(measureToCubeMappingInWidgets[measureName] ?? []), ...(measureToCubeMappingInDashboards ?? [])]);
  if(cubes.length > 0){
    acc[measureName] = cubes;
  }
  return acc;
}, {});
```

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  ...contentServerForTests.children!.ui.children,
+  calculated_measures: uiCalculatedMeasuresFolder,
+  dashboards: uiDashboardsFolder,
+  widgets: uiWidgetsFolder,
+};
+migrateCalculatedMeasures(contentServerForTests, dataModel);
+
+describe("migrateCalculatedMeasures", () => {
+  it("Mutates `contentServerForTests` by migrating calculated measures and removing calculated measure definitions from saved widgets and saved dashboards", () => {
+    expect(contentServerForTests).toMatchSnapshot();
```
antoinetissier, line 23: I don't think people would read the snapshot, it's really big.
The smaller unit tests are great and do the job, so let's remove this one.
We could check however that the `ui/calculated_measures` folder gets removed

antoinetissier (commented):
A few remarks but it looks good already

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.test.ts
```diff
@@ -0,0 +1,92 @@
...
+const dataModel = dataModelsForTests.sandbox;
+
+const contentServerForTests = _cloneDeep(contentServer);
+
+contentServerForTests.children!.ui.children = {
+  ...contentServerForTests.children!.ui.children,
+  calculated_measures: uiCalculatedMeasuresFolder,
+  dashboards: uiDashboardsFolder,
+  widgets: uiWidgetsFolder,
+};
```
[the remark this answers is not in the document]
RaspberryChaos: I did this because the `widgets` in the `contentServer` test resource contain no calculated measures, and the `dashboards` contain only one which is not in the `calculated_measures` folder. None of the `calculated_measures` from the `contentServer` test resource would have been migrated because they aren't used in any of the saved widgets or dashboards. So in order to test this function properly, I need the `uiWidgetsFolder`, `uiDashboardsFolder` and `uiCalculatedMeasuresFolder`. But maybe `_cloneDeep` is not actually necessary ?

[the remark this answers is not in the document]
antoinetissier: Makes sense.
Ideally, it would be nice if we could harmonize the test resources and keep only `contentServer`: keep only a minimal amount in there, but make sure that it covers the calculated measures as well.
It's not the priority though, so we can harmonize the test resources in a separate PR.
In the meantime, what you did is good.

antoinetissier (approved):
Looks great ! Let's just go back to the clone deep, and harmonize the test resources later.

PR #61, opened 2022-12-19
UI-7645 - Wire the migration frame

antoinetissier:
The `migrateXyz` functions are curried in `bin.ts` and passed down to each migration step.
The indexed `dataModels` are passed down to the callbacks.
`keysOfPluginsToRemove` is passed down to the `migrateDashboard` callback. 
Each callback is called within `produce`.
Their signature is adapted to reflect that (return type = `void`, allowing any mutation, or `XyzState` if the caller wants to return a completely new object).


src/cli/bin.ts
```diff
@@ -164,22 +148,67 @@ yargs
...
-      const toVersionIndex = toVersions.indexOf(toVersion);
+      const doesReportIncludeStacks = stack;
 
-      for (const { migrate } of migrationFunctions.slice(
-        fromVersionIndex,
-        toVersionIndex + 1,
-      )) {
-        await migrate(contentServer, {
-          counters,
+      // Handle the special case of 4.3 to 5.0 separately, as:
```
eliasinho, line 153: Hmm, I'm not entirely convinced by those two arguments:
- the fact that the content server structure is different doesn't keep us from composing `migrate_43_to_50` with the other migration functions. The logic to parse the legacy content server is inside `migrate_43_to_50` anyway.
- the difference is that we don't use the generic `migrateXyz` functions inside, and that we need the `servers`, which is reflected by the signature of the function being different. But it does not really keep us from composing `migrate_43_to_50` with the other migration functions, right? We could haver `servers` and the `migrateXyz` parameters as optional

So, unless I'm missing something, when I read this comment, I'm not entirely convinced that we **need** to handle this separately from the other migrations. I would adapt the comment, maybe saying more that it's cleaner for `migrate_43_to_50` to have its own signature and to not have mixed with the other functions signatures. Or even not handle the `4.3` case separately, if that's possible. What do you think?

Also, I would add the fact that `migrate_43_to_50` is asynchronous, as opposed to the other ones, because that definitely plays for handling this case separately, right?

antoinetissier: To sum up:
- The fact that the content server is completely different means that the `getMigrateXyz` functions don't apply.
- The `servers` are needed only in this function and not in the other ones.
- It is the only async one.

These are the arguments for not composing: we don't want to pollute all the other functions by making some arguments optional, or adding unnecessary ones.

I'll add the async argument in the comment 👍 

For the general shape of the comment I'd like it to remain as short and simple as possible. If you have an improvement in mind can you make a suggestion ?

src/5.0_to_5.1/migrateDashboard.ts
```diff
@@ -0,0 +1,12 @@
+import { DashboardState as DashboardState50 } from "@activeviam/activeui-sdk-5.0";
+import { DashboardState as DashboardState51 } from "@activeviam/activeui-sdk-5.1";
+import { MigrateDashboardCallback } from "../migration.types";
+
+export const migrateDashboard: MigrateDashboardCallback<
+  DashboardState50,
+  DashboardState51
+> = () => {
+  // TODO
```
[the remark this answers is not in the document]
antoinetissier: I'll make a PR right after this one that adds the migration of context values and filters, and then we can test from there.

src/_getMetaData.ts
```diff
@@ -0,0 +1,17 @@
+import { AMetaData, ContentRecord } from "@activeviam/activeui-sdk-5.0";
+
+/**
+ * Returns the metadata of a file under `structure`.
+ */
+export function _getMetaData<MetaData extends AMetaData = AMetaData>(
```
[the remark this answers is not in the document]
antoinetissier: Yes, and especially when it reveals a bugged implementation 😇

src/5.0_to_5.1/migrateDashboard.ts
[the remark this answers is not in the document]
eliasinho: Sounds good!

src/_getMetaData.ts
```diff
@@ -0,0 +1,17 @@
...
+  structure: ContentRecord,
+  pathToParentFolder: string[],
+  fileId: string,
+): MetaData {
+  const metadataRecord = [
+    ...pathToParentFolder,
+    fileId,
+    `${fileId}_metadata`,
+  ].reduce((acc, id) => acc.children![id], structure);
```
eliasinho, line 15: I feel like we're using the `!` with confidence, I'm not sure if we should worry about the case where a bug happened somewhere above, and `pathToParentFolder` is inconsistent with `fileId`. Should we throw maybe? The error could be cleanly caught and added to the error report in the end. What do you think? Maybe I'm overthinking this

src/cli/bin.ts
eliasinho: I think you could sum up the first two dots with sth like: `its function has a different signature from the other migration functions`. I find that bit more convincing.
I find the sentence "the legacy content server structure is completely different" confusing because each migration step has its own logic anyway, whereas the impact it has (less parameters needed in the signature), is the reason why we're handling it separately.

Feel free to reformulate, or to disagree of course!

src/getMigrateWidgets.ts
```diff
@@ -44,22 +48,53 @@ export const getMigrateWidgets =
...
+          fileId,
+          name: metadata.name!,
+        });
+        delete widgetsContent[fileId];
+        const parentFolder = folderId.reduce(
+          (acc, id) => acc.children![id],
+          widgetsStructure,
+        );
+        delete parentFolder.children![fileId];
+        continue;
```
eliasinho, line 85: Out of curiosity, why using this instead of having the rest of the migration in an `else` statement?
I admit that I'm not very familiar with `continue`

src/getMigrateWidgets.ts
```diff
@@ -44,22 +48,53 @@ export const getMigrateWidgets =
       contentServer.children?.ui.children?.widgets.children?.structure!;
     const filesAncestry = _getFilesAncestry(widgetsStructure);
 
+    const migrateWidget = produce(callback);
```
eliasinho, line 51: Just to make sure I understand, `produce` here returns a function that executes the callback when called, right?

src/migration.types.ts
```diff
@@ -75,22 +77,77 @@ export interface ErrorReport {
...
   dashboardState: FromDashboardState,
-) => ToDashboardState;
+  {
+    dataModels,
+    keysOfWidgetPluginsToRemove,
+  }: {
+    dataModels: { [serverKey: string]: DataModel };
+    keysOfWidgetPluginsToRemove: string[];
+  },
+) => void | ToDashboardState;
```
eliasinho, line 91: Why the union of the return types?

src/_getMetaData.ts
```diff
@@ -0,0 +1,17 @@
+import { AMetaData, ContentRecord } from "@activeviam/activeui-sdk-5.0";
+
+/**
+ * Returns the metadata of the file with id `fileId` under `structure`.
+ */
+export function _getMetaData<MetaData extends AMetaData = AMetaData>(
```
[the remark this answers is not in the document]
antoinetissier: The naming is inline with the type and with the concept.
The file itself is referred to as the "metadata record" or "metadata file" in the monorepo.

src/_getMetaData.ts
```diff
@@ -0,0 +1,17 @@
...
+  structure: ContentRecord,
+  pathToParentFolder: string[],
```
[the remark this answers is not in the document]
antoinetissier: We'd still have to do `_last` to get the id of the file itself, and to use it for the id of the medata file.
Keeping the signature like this makes it easier to wire it by the caller, and it's explicit.

src/_getMetaData.ts
antoinetissier: I don't mind the idea, although I would tend to throw higher up if there is a structural issue: at this stage you expect the `pathToParentFolder` to have been computed by a first pass within the structure anyway.
I would rather not complexify the PR for now to unblock the rest.

src/getMigrateWidgets.ts
antoinetissier: It's just a coding style preference, but it's one that we try to follow in the monorepo: avoid nesting, and deal with edge cases with early returns (or `continue`/`break`, etc.) as soon as possible.

src/getMigrateWidgets.ts
antoinetissier: Yes, it's the curried form of `produce`.

src/migration.types.ts
antoinetissier: To align on the types of the recipe `produce`: you can either mutate, or return a completely new object.
I think it makes sense even if the caller doesn't know that it will ultimately be used in `produce`.

antoinetissier:
> I find the sentence "the legacy content server structure is completely different" confusing because each migration step has its own logic anyway

Yes but in the 5.x to 5.y the assumption is that they all share the same content server structure, and that we're mostly just migrating the content of the files.

I can attempt to reword though.

src/getMigrateWidgets.ts
eliasinho: Okay, noted!

src/_getMetaData.ts
eliasinho: Okay, I'm fine with not doing this here!

eliasinho (commented):
Thanks for updating the comment! I'm just a bit confused with the typing we're going for now, but appart from that, everything looks great!

antoinetissier:
> migratedDashboard = migrateDashboard(...)

To be more exact it boils down to `migratedDashboard = produce(callback)(...)`.
So even if you return `void` and just perform mutations in the `callback`, in the end you get a new immutable object.

See the [immer doc](https://immerjs.github.io/immer/curried-produce/).

eliasinho:
> > migratedDashboard = migrateDashboard(...)
> 
> To be more exact it boils down to `migratedDashboard = produce(callback)(...)`. So even if you return `void` and just perform mutations in the `callback`, in the end you get a new immutable object.
> 
> See the [immer doc](https://immerjs.github.io/immer/curried-produce/).

Okay, it makes sense now, thanks!

eliasinho (approved):
Looks good to me! ✅

PR #62, opened 2022-12-20
UI-7645 - Bump to allow usage of _removeWidgetFromPage

antoinetissier:
See the new util added in https://github.com/activeviam/activeui/pull/3139, that we need to leverage here.
With the recent introduction of the first ESM package in the monorepo (`@activeviam/utils`), we need to change the Jest file transform and an experimental flag to make Jest work.
See https://jestjs.io/fr/docs/ecmascript-modules


eliasinho (commented):
Looks good to me! You're going to actually import `_removeWidgetFromPage` in a following PR then?

Just waiting for the build to fixed before approving.

antoinetissier:
> You're going to actually import _removeWidgetFromPage in a following PR then?

Yes

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.test.ts
```diff
@@ -1,12 +1,12 @@
 import { migrateCalculatedMeasures } from "./migrateCalculatedMeasures";
-import { dataModelsForTests } from "@activeviam/data-model-5.0";
+import { sandboxDataModel } from "@activeviam/data-model-5.1/dist/__test_resources__";
```
eliasinho, line 2: Can't we export that from `@activeviam/test-utils` instead? Just the export, not the definition. 
As `@activeviam/test-utils` is a dev dependency of the app, it wouldn't increase the bundle size, and it would avoid having this import in `*/dist/__test_resources__`

Or we consider that [UI-8016](https://support.activeviam.com/jira/browse/UI-8016) will remove the `dist` and that'll be enough.

But as we're in a separate repo, I find it cleaner to not have `__test_resources__` at all. Also I like the idea of having test resources exported from `@activeviam/test-utils`: `@activeviam/test-utils` would behave a bit like `activeui-sdk-scripts` and be the point of entry for test related resources imports.

What do you think? 

I guess it needs to be a wider discussion, and I know what you've done is the current convention, so if you disagree feel free to leave this as it is.

antoinetissier: Yes I prefer to leave this for outside this PR.
And in any case I'm against putting everything in `@activeviam/test-utils`:
- we'll end up having circular dependencies
- we should always keep the test resources that originate from a package in the package itself (e.g. a datamodel snapshot belongs in `data-model`, not `test-utils`)

eliasinho: Okay, I'll approve the PR then.

> And in any case I'm against putting everything in @activeviam/test-utils

But just to be clear, I'm not talking about putting everything in `@activeviam/test-utils`. I'm saying that for test resources that need to be used outside of the monorepo:
  -  leave their definition in their origin package
  - also export them from `@activeviam/test-utils`
  - inside the monorepo, import their origin package
  - from outside the monorepo, use `@activeviam/test-utils`

The bottom line of this is to have a clearer import when needing to use test resources from outside the monorepo.
There might be some circular dependencies I'm not foreseeing though, indeed.

eliasinho (approved):
Looks good to me!

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.test.ts
antoinetissier: Ah I see, so in:

> @activeviam/test-utils would behave a bit like activeui-sdk-scripts and be the point of entry for test related resources imports.

you meant `activeui-sdk` rather than `activeui-sdk-scripts` right ?

If so yeah no problem with that 👍

eliasinho: Ah yes, my bad!!

PR #64, opened 2022-12-20
UI-7645 - Logic to migrate widgets within dashboards

antoinetissier:
Adding a function that can be called from the callback passed to `migrateDashboards`, to migrate all widgets within the dashboard.
Essentially the structure is the following:
```
// In bin.ts
const migrateDashboards = getMigrateDashboards(...);
const migrateWidgets = getMigrateWidgets(...);

// In the X_To_Y migration folder
const migrateWidget = () => {
  // some widget specific logic
}

const migrateDashboard = () => {
  // some dashboard specific logic
  migrateWidgetsWithinDashboard(migrateWidget );
}

const migrateXToY = ({migrateDashboards, migrateWidgets}) => {
  migrateDashboards(migrateDashboard);
  migrateWidgets(migrateWidget);
}
```

There is an example of this [in the next PR](https://github.com/activeviam/activeui-migration/pull/65/files#diff-758cb663fecc1f87f4acb907c5f39976afc64a6a3c4c424f9b4ed9162b825520).


package.json
```diff
@@ -31,12 +31,13 @@
   },
   "dependencies": {
     "@activeviam/activeui-sdk-5.0": "npm:@activeviam/activeui-sdk@5.0.15",
-    "@activeviam/activeui-sdk-5.1": "npm:@activeviam/activeui-sdk@5.1.0-20221220092919-bb816dc",
+    "@activeviam/activeui-sdk-5.1": "npm:@activeviam/activeui-sdk@5.1.0-20221220105623-9a2eef6",
```
antoinetissier, line 34: I just realized that I had actually taken the parent commit and not the ocmmit itself 🤦 
At least this allowed to isolate the changes necessary to make Jest work with ESM packages.

eliasinho: Shouldn't it be [this commit](https://github.com/activeviam/activeui/commit/15897ed90c038f5a452548b5561ac6dae38af4c8) though: 15897ed90c038f5a452548b5561ac6dae38af4c8?

src/migrateWidgetsWithinDashboard.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Mutates `dashboardState` by calling `callback` on each underlying widget state.
+ */
+export function migrateWidgetsWithinDashboard<
+  FromWidgetState extends { widgetKey: string; name: string },
+  FromDashboardState extends {
+    pages: {
+      [pageKey: string]: {
+        name: string;
```
eliasinho, line 15: Could we use `DashboardPageState` here?

src/migrateWidgetsWithinDashboard.ts
```diff
@@ -0,0 +1,67 @@
+import { DataModel, Layout, getLayoutPath } from "@activeviam/activeui-sdk-5.1";
+import { _removeWidgetFromPage } from "@activeviam/dashboard-base-5.1";
+import _forEach from "lodash/forEach";
+import { MigrateWidgetCallback } from "./migration.types";
+import { WidgetFlaggedForRemovalError } from "./WidgetFlaggedForRemovalError";
+
+/**
+ * Mutates `dashboardState` by calling `callback` on each underlying widget state.
+ */
+export function migrateWidgetsWithinDashboard<
```
eliasinho, line 10: I think it's good if we have a unit test for this function too.
I can handle it in a separate PR if you prefer.

eliasinho (commented):
Looks great! I just have a few remarks/questions, but I like the abstraction of this, with both the currying of the callback and of `onError` in `migrateWidgetsWithinDashboard`.

In terms of naming, we have `getMigrateWidgets` that handles the saved widgets only. In `4.3_to_5.0`, `migrateWidget` seems to be taking care of all widgets. 
I mean there is nothing to change in this PR I think, because you already went for `migrateWidgetsWithinDashboard`, which I like. But with the currying there is going to be a lot of `migrateXyz` functions defined here and there, so I just want to make sure that first I understand the specificity, and that we don't introduce any confusion.

antoinetissier:
> Shouldn't it be [this commit](https://github.com/activeviam/activeui/commit/15897ed90c038f5a452548b5561ac6dae38af4c8) though: 15897ed90c038f5a452548b5561ac6dae38af4c8?

I took the most recent one on `main`, so we gain a few commits as a nice to have.

src/migrateWidgetsWithinDashboard.ts
antoinetissier: We could, and same for `DashboardState` and `WidgetState`, but it would tie it down to a specific version of the type (e.g. the 5.0 one).
Here I wanted to avoid that since `migrateWidgetsWithinDashboard` will be called by all `migrateXToY` functions.
And the content type is not needed to implement the iteration logic: just this shell is enough.

It's not a very strong preference though, so if you prefer to import from `5.0` and try to type more precisely I don't mind giving it a go.

antoinetissier:
See the snippet in the [PR description](https://github.com/activeviam/activeui-migration/pull/64#issue-1504612784): in the AcitveUI migrations `migrateWidget` ends up being called both for saved widgets and widgets within dashboards, like in 4.3 to 5.0:
- `migrateWidgets` (notice the plural) only deals with saved widgets, and for each saved widget it applies the `migrateWidget` callback
- `migrateDasboards` calls `migrateDashboard` for each saved dashboard. It is up to each implementation of `migrateDashboard` to call `migrateWidgetsWithinDashboard` if necessary, with `migrateWidget` as its callback.

src/migrateWidgetsWithinDashboard.ts
antoinetissier: I added a quick one, but we could be more thorough in further PRs when focusing on testing.

eliasinho (approved):
Thanks for adding the test!

> migrateWidgets (notice the plural) only deals with saved widgets, and for each saved widget it applies the migrateWidget callback

Yes, that's the thing, `migrateWidgets` doesn't reflect the fact that it only takes care of saved widgets, and it's misleading IMHO. I think we should rename it to `migrateSavedWidgets`.
It's out of scope for this PR anyway, we can talk about it later.

> It's not a very strong preference though, so if you prefer to import from 5.0 and try to type more precisely I don't mind giving it a go.

No, I don't feel strongly either, and I'm fine actually with this approach!

▶ Looks good to me! ✅

PR #65, opened 2022-12-20
UI-7645 - Start implementing migration from 5.0 to 5.1

antoinetissier:
Starting the actual migration script implementation so that we can start testing the CLI on real cases.


src/5.0_to_5.1/migrateContextValues.ts
```diff
@@ -0,0 +1,11 @@
+import { QueryContextEntry } from "@activeviam/activeui-sdk-5.0";
+
+/**
+ * Stringifes the values of the entries in `queryContext`.
+ * Mutates `queryContext`.
+ */
+export function migrateContextValues(queryContext?: QueryContextEntry[]): void {
```
eliasinho, line 7:
```suggestion
export function migrateContextValues(queryContext: DashboardState50["queryContext"] = []): void {
```
There are two suggestions here:
- using `DashboardState50["queryContext"]` instead of `QueryContextEntry[]`: I find that weird for a function to accept a single parameter as being optional. As I read it in its current state, the role of this function is to migrate the values of a query context, so if there is no queryContext, there's no point in calling the function at all. However this function will always be applied to the `queryContext` part of the dashboard state, which will always be typed as potentially undefined, so I agree it's better reflected and handled directly here. And I find this better reflected typed with `DashboardState50["queryContext"]`.
- defaulting the value as an empty array. That way `queryContext` is always an array and you wouldn't use `?.` after `queryContext`, which I find more correct. But honestly it's more of a preference than something I feel strongly about, so feel free not to apply this suggestion.

src/5.0_to_5.1/migrateFilters.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Wraps the MDX of each filter in `filters`.
+ * Mutates `filters`.
+ */
+export function migrateFilters(filters: any[] | undefined): void {
```
eliasinho, line 5: Typing this as `any` is the downside of the choice of going for mutations everywhere, right?
Can we add a comment explaining why we're using `any`?

src/5.0_to_5.1/migrateFilters.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Wraps the MDX of each filter in `filters`.
```
eliasinho, line 2:
```suggestion
 * Wraps the MDX of each filter in `mdx`.
```
Right?

eliasinho (commented):
I would add the classic comment of adding unit tests for `migrateContextValues` and `migrateFilters`. But again, we take care of that in a further PR if you prefer.

src/5.0_to_5.1/migrateFilters.ts
antoinetissier: Hmm no that's what I meant: each element of `filters` is a filter.
In 5.0 it is represented by raw mdx, and in 5.0 it must have the form `{mdx}`

src/5.0_to_5.1/migrateContextValues.ts
antoinetissier: I find the second option misleading because it applies not just to the state of a dashboard, but also of a widget and a page.
And doing `DashboardState50["queryContext"] | DashboardPageState50["queryContext"] | AWidgetState50["queryContext"]` looks like turning something simple into something complicated.
I think this is a good reflection of why, in general, variables should be named and typed after their nature rather than after how they're used.

The second option would mean adding an unnecessary empty array in the saved content.
It's not a huge deal but if we do that for many pieces of state and for 10 000 saved items it's a shame.

  With that in mind, having an optional argument is acceptable imo.

src/5.0_to_5.1/migrateFilters.ts
antoinetissier: Yes it it is the downside of the mutations indeed.
I think something btter than `any` could be to type the input as `Mdx[]`, and then do `ts-expect-error` when mutating.
It works realtively well for small logic like this where it's basically a one-liner, however it's more annoying when the logic grows bigger: it's annoying to have to call call `ts-expect-error` for every single mutation.

package.json
```diff
@@ -26,6 +26,9 @@
     "test": "cross-env NODE_OPTIONS='--experimental-vm-modules' activeui-sdk-scripts test --config jest.config.js",
     "ci:test": "cross-env CI=true yarn test"
   },
+  "eslintConfig": {
+    "extends": "activeui"
+  },
```
antoinetissier, line 31: This is to allow `@ts-expect-error` when given an explanation.

src/5.0_to_5.1/migrateContextValues.ts
eliasinho: Yeah, you're raising good points I hadn't thought about. I'm happy to keep it this way, it's a detail anyway.

I'd just like your opinion: in general, don't you think it's weird to have such a parameter of a function to be optional? Like here the only role of the function is to mutate its parameter. Like there is in theory no point in calling the function at all if this parameter is not here. To me, it's not the role of the function to handle the case where its parameter would be undefined, it shouldn't care about it. Is it just me who is nitpicky about this or do you agree (again, I'm talking in a general way, not in this specific case)?

src/5.0_to_5.1/migrateFilters.ts
eliasinho: Aaah, okay, I misunderstood the sentence. I find the comment confusing, but I don't have a clear idea of a good improvement for it. Maybe this?
```suggestion
 * Wraps each filter in `filters` in an object.
```

Also I don't see the need for "the MDX of", each 5.0 filter is an MDX

src/5.0_to_5.1/migrateContextValues.test.ts
```diff
@@ -0,0 +1,10 @@
+import { migrateContextValues } from "./migrateContextValues";
+
+describe("migrateContextValues", () => {
+  it("stringifies the query context values", () => {
+    const contextValues = [{ key: "queriesTimeLimit", value: 30 }];
+    migrateContextValues(contextValues);
+    expect(contextValues).toHaveLength(1);
```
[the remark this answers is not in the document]
antoinetissier: The fact that the values are stringified is in the test description already, but okay for `toStrictEqual`

src/5.0_to_5.1/migrateContextValues.ts
antoinetissier: It depends a bit on the usage imo:  the pro is to factorize the handling of the edge case, and in simple cases like this I think pragmatically it's a good enough argument.
But when the call stack grows big, then I think it's wrong to delegate the handling of the edge case several times down the call stack => edge cases should be handled as soon as possible.

So in general I agree yes, and in fact I find it very frustrating that you don't have another choice than the optional arg pattern with React hooks for example.

src/5.0_to_5.1/migrateFilters.ts
```diff
@@ -0,0 +1,12 @@
+import type { Mdx } from "@activeviam/activeui-sdk-5.0";
+
+/**
+ * Wraps the MDX of each filter in `filters`.
+ * Mutates `filters`.
+ */
+export function migrateFilters(filters?: Mdx[]): void {
```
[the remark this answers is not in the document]
antoinetissier: 👀 Very good point, the deserialization/serialization is missing in the current implementation. I'll add it 👍

eliasinho:
@Nouzbe this PR does two things:
- it introduces the `migrateFilters` and `migrateContextValues` 5.0 --> 5.1 migration functions. It triggers them in the `migrateDashboard` and `migrateWidget` functions.
- it introduces a deserialization/serialization mechanism: to manipulate more easily the content objects, we thought it would be better to deserialize them at first.

There are two problems with the deserialization/serialization mechanism:
- the deserialization mechanism from AUI-5 [uses `produce` from `immer`](https://github.com/activeviam/activeui/blob/main/packages/widget/src/utils/deserializeWidgetMapping.ts#L35), hence returns immutable objects. It causes us not to be able to directly mutate the deserialized states. That's why we moved `produce` from wrapping the callback in the `getMigrateXyz` functions to being used at a lower level, inside the `migrateXyz` functions, after the deserialization. An alternative we also thought of but liked less:
  - add an `autoFreeze` boolean to the AUI deserialization functions, using [this immer concept](https://immerjs.github.io/immer/freezing/). Would require some investigation of how it works though
- as `migrateWidget` starts by deserializing the widget state, it can't be called like this in `migrateWidgetsWithinDashboard` anymore, because the passed widget state is already deserialized at this point. Hence why we introduced `mutateWidget`, which triggers all the widget state modifications after deserialization, and be reused in `migrateWidgetsWithinDashboard`. For consistency, we introduced `mutateDashboard` too. Alternatives we also thought of but liked less:
  - passing an `doesRequireToDeserialize` boolean to `migrateWidget` --> doesn't look clean at all
  - serializing the widget state before `migrateWidgetsWithinDashboard` is called --> causes extra useless serialization+deserialization step

There's a couple other things that I noticed we forgot to do in this PR:
- `migrateFilters` shouldn't have been introduced that way. It should have been called `migrateFilter`, used in a `forEach` in `migrateDashboard` and `migrateWidget`, as well as passed in a callback to the function returned by `getMigrateSavedFilters`.
- after the introduction of `mutateWidget`, `migrateWidget` will only be used for saved widgets, so it should be renamed `migrateSavedWidget`
- we might want to have the same deserialization/serialization mechanism for filters too, using `parse`/`stringify`.

antoinetissier:
> migrateFilters shouldn't have been introduced that way. It should have been called migrateFilter, used in a forEach in migrateDashboard and migrateWidget, as well as passed in a callback to the function returned by getMigrateSavedFilters

I disagree with this because saved filters have a different form than filters within a widget/dashboard state.

src/getMigrateDashboards.ts
```diff
@@ -42,7 +43,14 @@ export const getMigrateDashboards =
   ) =>
   <FromDashboardState, ToDashboardState>(
     callback: MigrateDashboardCallback<FromDashboardState, ToDashboardState>,
+    deserialize: (state: FromDashboardState) => FromDashboardState,
+    serialize: (state: ToDashboardState) => ToDashboardState,
   ): void => {
+    // This function returned by `getMigrateDashboards` and accessing its outer scope variable `contentServer` forms a closure.
+    // It causes some parts of the `contentServer` object to not be writable, hence not mutable.
+    // This is done to override that behavior and have a fully mutable `contentServer` object.
```
antoinetissier, line 51: What causes the immutability ? Can't we use `produce` in the right places instead ?
Also, I'm not sure I get the link between the function closure and the immutability.

src/getMigrateDashboards.ts
```diff
@@ -91,11 +98,20 @@ export const getMigrateDashboards =
...
+        const deserializedMigratedDashboard = produce(
+          deserializedDashboard,
+          (draft) =>
+            callback(draft as FromDashboardState, {
+              dataModels,
+              keysOfWidgetPluginsToRemove,
+              onErrorWhileMigratingWidget,
+            }),
+        );
+        // At the end of the migration, `deserializedMigratedDashboard` is of type `ToDashboardState`.
```
antoinetissier, line 111: Let's put this line right above the cast

src/getMigrateSavedFilters.ts
```diff
@@ -39,25 +39,28 @@ export const getMigrateSavedFilters =
   <FromFilterState, ToFilterState>(
     callback: MigrateFilterCallback<FromFilterState, ToFilterState>,
   ): void => {
+    // This function returned by `getMigrateDashboards` and accessing its outer scope variable `contentServer` forms a closure.
```
antoinetissier, line 42:
```suggestion
    // This function returned by `getMigrateFilters` and accessing its outer scope variable `contentServer` forms a closure.
```

src/getMigrateSavedWidgets.ts
```diff
@@ -41,15 +42,20 @@ export const getMigrateSavedWidgets =
   ) =>
   <FromWidgetState, ToWidgetState>(
     callback: MigrateWidgetCallback<FromWidgetState, ToWidgetState>,
+    deserialize: (state: FromWidgetState) => FromWidgetState,
+    serialize: (state: ToWidgetState) => ToWidgetState,
   ): void => {
+    // This function returned by `getMigrateDashboards` and accessing its outer scope variable `contentServer` forms a closure.
```
antoinetissier, line 48:
```suggestion
    // This function returned by `getMigrateWidgets` and accessing its outer scope variable `contentServer` forms a closure.
```

src/5.0_to_5.1/migrateSavedWidget.ts
```diff
@@ -0,0 +1,29 @@
...
+  serializeWidgetState,
+} from "@activeviam/activeui-sdk-5.1";
+import { MigrateWidgetCallback } from "../migration.types";
+import { produce } from "immer";
+import { mutateWidget } from "./mutateWidget";
+
+/**
+ * Returns the 5.1 migrated version of the 5.0 `serializedWidgetState`.
+ * Used only for saved widgets, not for widgets within dashboards.
+ */
```
antoinetissier: If we go for `serialize`/`deserialize` at a higher level in `getMigrateXyz`, then this comment and the implementation with the serialized state doesn't work anymore.
Same for the other `migrateXyz` functions.

src/5.0_to_5.1/migrateSavedFilter.ts
```diff
@@ -0,0 +1,12 @@
...
+import { MdxString } from "@activeviam/mdx-5.0";
+
+/**
+ * Returns the 5.1 migrated version of the 5.0 `filterState`.
+ * Used only for saved filters, not for filters within dashboards.
+ */
+export const migrateSavedFilter: MigrateFilterCallback<
+  MdxString,
+  FilterState51<"serialized">
+> = (filterState) => ({ mdx: filterState });
```
antoinetissier, line 12: There is no need to migrate saved filters: if you look at what is saved in the content server it already has the wanted format.
image
So let's revert this commit.

src/getMigrateDashboards.ts
```diff
@@ -42,7 +43,14 @@ export const getMigrateDashboards =
   ) =>
   <FromDashboardState, ToDashboardState>(
     callback: MigrateDashboardCallback<FromDashboardState, ToDashboardState>,
+    deserialize: (state: FromDashboardState) => FromDashboardState,
+    serialize: (state: ToDashboardState) => ToDashboardState,
```
antoinetissier, line 47: The input types are not accurate: theoretically it should be `From/ToDashboardState<serialized>`.
The input types are generic themselves so we cannot do that exactly, but we could use `unknown` (or `any` exceptionally if `unknown` causes too many issues) to avoid being mislading.

src/getMigrateDashboards.ts
antoinetissier: As discussed together this is probably because `migrateCalculatedMeasures` is called before and creates an immutable object in the first place.
https://github.com/activeviam/activeui-migration/pull/70 will lift this problem.

src/5.0_to_5.1/migrateSavedFilter.ts
antoinetissier: While the implementation needs to change (there is nothing to do), we might as well keep a dummy `migrateSaveFilter` just for the sake of setting a precedent, and benefitting from the changes to `parse` and `stringify` for compound identifiers that occurred between 5.0 and 5.1.

src/getMigrateDashboards.ts
```diff
@@ -41,25 +41,26 @@ export const getMigrateDashboards =
     },
   ) =>
   <FromDashboardState, ToDashboardState>(
+    deserialize: (state: any) => FromDashboardState,
```
eliasinho, line 44: To avoid using `any`, we could have more type parameters, right?
Add `SerializedFromDashboardState` and `SerializedToDashboardState`, and modifying `FromDashboardState` --> `DeserializedFromDashboardState` and `ToDashboardState` --> `DeserializedToDashboardState`

src/getMigrateSavedFilters.ts
```diff
@@ -37,31 +37,44 @@ export const getMigrateSavedFilters =
...
-      const metadata = _getMetaData(filtersStructure, folderId, fileId);
+      const metadata = _getMetaData(structure, folderId, fileId);
 
       try {
-        migratedFilter = migrateFilter(filter, { dataModels });
-        // The filter was fully migrated.
+        const deserializedFilter = deserialize(filter);
+        const deserializedMigratedFilter = produce(
+          deserializedFilter,
+          (draft) => callback(draft as FromFilterState, { dataModels }),
```
eliasinho, line 65: Why do we need to cast here? If `deserialize` returns `FromFilterState`

eliasinho (commented):
Looks really good! I just have two remarks.

src/getMigrateDashboards.ts
antoinetissier: We could, but I decided against it for the following reasons:
- It would double the amount of generics without adding more type safety, as ultimately `any` is used anyway, and we're doing mutations.
- Having `FromSerializedDashboardState` and `FromDashboardState` as two different generics would not be aligned with `DashboardState<"serialized">` and `DashboardState`, where the serialization is differentiated with a generic in one same type, as opposed to having two different types.

Imo the `MigrationFunction` types there are helpful to give a structure and an outline to make the code easier to read, but not to ensure a 100% type safety.

src/getMigrateSavedFilters.ts
antoinetissier: `callback` expects a `FromFilterState`, but `TypeScript` complains that it receives a `Draft<FromFilterState>`. I think `immer` imposes that to warn against mutability, but here we know what we're doing so it's not a problem.
We could probably type `callback` as a `Recipe` instead, but that would make it harder to read for someone just wanting to write a script, and couple our types to `produce`.
So I prefer to cast and keep the simple "mutate the input" contract of `migrateXyz`.

src/getMigrateDashboards.ts
eliasinho: > Having FromSerializedDashboardState and FromDashboardState as two different generics would not be aligned with DashboardState<"serialized"> and DashboardState

I don't see that as a concern we should have. To me `@activeviam/activeui-sdk` is a different entity as `activeui-migration`, and we don't necessarily have in `activeui-migration` to follow the types structure that's defined in `@activeviam/activeui-sdk`. Especially if we need a different structure for a good reason.

>  imo the MigrationFunction types are mostly helpful to give a structure and an outline to make the code easier to read, but we cannot ensure a 100% type safety anyway.

> It would double the amount of generics without adding more type safety

I completely agree with those statements though. 

I'm leaning towards introducing the `FromSerializedXyzState` and `ToSerializedXyzState` types. I really think it would make the code easier to read, the currying structure with the generic serialization/deserialization making it pretty abstract. Especially with the order of the parameters that you went for, it would make the three steps (`deserialize`, `callback`, `serialize`) really clear when reading the function signature. 

It does double the amount of state types we use for those generic functions, but I think 4 type generics for the `getMigrateXyz` is still legible. It makes the ones of `MigrationFunction` go up to 12, which is a lot, but I noticed when used, `MigrationFunction` shouldn't need any type parameters, so I'm not concerned about it getting too heavy.

Are you happy to go for it?

On the same topic, I have an extra comment. I noticed we use `any` as the second type parameter of `callback` in [the signature of `migrateWidgetsWithinDashboard`](https://github.com/activeviam/activeui-migration/blob/main/src/migrateWidgetsWithinDashboard.ts#L23). I think we can introduce a `ToDashboardState` generic here too, to avoid using `any`, which makes the code less clear IMO.

What do you think?

src/getMigrateSavedFilters.ts
eliasinho: Yeah, I agree with you, casting sounds better.
I was gonna ask you to add a comment above to explain why we need to do it, but actually I'm not sure we need to. It makes sense that `draft` needs to be a `FromFilterState` at this point. So feel free to add one if you feel the need, but otherwise I'm happy to keep this as it is.

src/getMigrateDashboards.ts
antoinetissier: I pushed some commits with these changes: as expected it's more verbose, but I don't mind keeping it this way if you think it adds readability.
We could also keep just part of it: for example imo adding the generics to `migrate_5.0_to_5.1` does not add readability not more type safety than what we already had, so we could remove that part.

eliasinho: Yeah, I agree we could not add the generics in `migrate_50_to_51`, at least for now. It would enlighten the function definition, and if we ever notice that it's better to have them, for whatever reason, we can add them back very quickly

But honestly I don't think I feel strongly about either options in this specific case, so as you already added them, feel free to leave this as it is.

src/migrateWidgetsWithinDashboard.ts
```diff
@@ -52,14 +53,13 @@ export function migrateWidgetsWithinDashboard<
...
         }
-
         callback(widgetState, { dataModels });
       } catch (e) {
         onError(e, {
           pageKey,
           leafKey,
           pageName: pageState.name,
-          widgetName: widgetState.name,
+          widgetName: widgetState.name ?? "",
```
eliasinho, line 62: Is it worth having a different default string here, like `"[This widget has no name]"`, to be a bit more explicit than a blank?  You could argue that it's the user's fault if they didn't name their widget, but I doesn't cost much to be slightly more explicit to avoid any confusion

eliasinho (approved):
There's my comment [here](https://github.com/activeviam/activeui-migration/pull/65#discussion_r1064796955), and a very small suggestion.

Whether you add the changes or not, I'm happy with the state of the PR, so I'll approve it already, so that can merge whenever you're done with the changes!

🎉🎉

src/migrateWidgetsWithinDashboard.ts
antoinetissier: It might be on purpose that the widget does not have a name, so I prefer to keep it this way.

PR #66, opened 2022-12-20
UI-7384 - Add error to `ErrorReport` if calculated measure is not migrated

RaspberryChaos:

If calculated measures are not used in any saved dashboards or widgets, they are not migrated from `5.0` to `5.1`. In this case a warning is added to the `ErrorReport`.


src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.ts
```diff
@@ -92,7 +102,27 @@ export function migrateCalculatedMeasures(
...
+        // `legacyCalculatedMeasuresFolder.children` contains `content` and `structure` properties.
+        const filesAncestry = _getFilesAncestry(
+          legacyCalculatedMeasuresFolder?.children?.structure!,
+        );
+        const calculatedMeasureId = calculatedMeasureNamesAndIds[measureName];
+        const folderId = filesAncestry[calculatedMeasureId].map(({ id }) => id);
+        const folderName = filesAncestry[calculatedMeasureId].map(
+          ({ name }) => name,
+        );
+        _addErrorToReport(errorReport, {
```
antoinetissier, line 114: Let's incremented the `removed` counter in this case.
In the other cases, let's increment the corresponding counter as well (`failed` or `success`).

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.ts
```diff
@@ -47,8 +51,14 @@ export function migrateCalculatedMeasures(
...
     ? _mapKeys(
         legacyCalculatedMeasuresFolder.children.content.children,
-        (value, key) =>
-          getCalculatedMeasureName(legacyCalculatedMeasuresFolder, key),
+        (value, key) => {
+          const calculatedMeasureName = getCalculatedMeasureName(
+            legacyCalculatedMeasuresFolder,
+            key,
+          );
+          calculatedMeasureNamesAndIds[calculatedMeasureName] = key;
```
antoinetissier, line 59: The starting point is the id.
So I'd find it more logical to have a single map where all the information is stored:
```
const namesOfCalculatedMeasuresToMigrate: string[] = [];
 const legacyCalculatedMeasures: {
    [id: string]: {record: ContentRecord, name: string};
  } = {}
_forEach(legacyCalculatedMeasuresFolder.children.content.children, (record, id) => {
 const name = getCalculatedMeasureName(legacyCalculatedMeasuresFolder, id);
  namesOfCalculatedMeasuresToMigrate.push(name);
  legacyCalculatedMeasures[id] = {name, record};
})
```

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.ts
```diff
@@ -124,6 +158,8 @@ export function migrateCalculatedMeasures(
             record;
         }
       });
+      // The calculated measure was successfully migrated.
+      counters.calculated_measures.success++;
```
antoinetissier, line 162: We should wrap all the block starting with `migrateCalculatedMeasureContent` with a `try/catch`.
If there is an error, then we increment the failures counter and add a the error to the report.

You will need the `filesAncestry`, `folderName` and `folderId` in this case as well, so you might want to define higher up than they currently are.

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasures.ts
```diff
@@ -88,42 +101,84 @@ export function migrateCalculatedMeasures(
...
+        // The calculated measure was not migrated.
+        counters.calculated_measures.failed++;
+        _addErrorToReport(errorReport, {
+          contentType: "calculated_measures",
+          folderId,
+          folderName,
+          fileErrorReport: {
+            error: {
+              message: `Warning: Calculated measure "${measureName}" was not migrated because it is not currently used in any saved widgets or dashboards.`,
+            },
```
antoinetissier, line 176: This warning message is not appropriate anymore in the case of a generic error.
Rather, we must take the error itself.
```suggestion
            error: _serializeError(error, {
              doesReportIncludeStacks,
            }),
```

antoinetissier (approved):
Just one last remark to take into account, but it looks good otherwise, I am approving ✅

PR #67, opened 2022-12-20
UI-7645 - Support several servers when migrating calculated measures

antoinetissier:
Aligning on the other migration functions to use `dataModels` instead of `dataModel`.
Indeed, there can be several target ActivePivot servers.

### DIFF
--- a/src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.ts
+++ b/src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.ts
@@ -6,6 +6,7 @@ import {
   isWidgetWithQueryState,
   MdxSelect,
   WidgetWithQueryState,
+  MdxDrillthrough,
 } from "@activeviam/activeui-sdk-5.0";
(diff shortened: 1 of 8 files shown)


RaspberryChaos (approved):
Looks good to me! ⭐

src/5.0_to_5.1/calculated-measures/migrateCalculatedMeasuresInWidgets.ts
```diff
@@ -34,50 +35,55 @@ export const migrateCalculatedMeasuresInWidgets = (
...
-                  ? _uniq([
-                      ...measureToCubeMapping[calculatedMeasureName],
-                      cubeName,
-                    ])
-                  : [cubeName];
-            },
-          );
-          const updatedWidgetState = produce(
+        if (
+          !isWidgetWithQueryState<MdxSelect | MdxDrillthrough>(
```
antoinetissier, line 46: In this file the diff seems big but it's just because I moved a branch to an early return.
The logic does not change more than wiring `dataModels` and `serverKey`, like in the other files.

PR #69, opened 2022-12-28
UI-7945 - Make the errors reporting generic and cross migrations

eliasinho:
## Description

This ticket is a prerequisite to [UI-7945](https://support.activeviam.com/jira/browse/UI-7945). The CLI should accept an `on-error` (name to challenge) optional argument that decides between the following behaviors:
* `"keep-original"`: if anything goes wrong during the migration of an item, the CLI `from-version` of the item is kept.
* `"keep-latter"`: if anything goes wrong during the migration of an item between n-version and n+1-version, the n-version of the item is kept.
* `"keep-going"`: if anything goes wrong during the migration of an item, keep trying to migrate it, even though it's corrupt, until the CLI `to-version`.
* any other behavior?

Note that an item here is either:
- a saved filter
- a saved widget
- a calculated measure
- a widget within a dashboard
- a dashboard without its widgets (dashboard or pages specific parts of a dashboard state)

That means if something is wrong with the migration of a widget within a dashboard, the option will only apply to that widget migration and not to the rest of the dashboard state. In such a case, the `counters.dashboards.partial` variable is incremented.

## Review

Aside from introducing the new CLI option, the main goal of this PR is to have the frame for handling the errors challenged:
- CLI option name
- behavior names
- the behaviors themselves


src/cli/bin.ts
```diff
@@ -117,6 +122,14 @@ yargs
         default: false,
         desc: "Whether stacktraces are included in the error report file.",
       });
+      args.option("on-error", {
+        alias: "e",
+        type: "string",
+        demandOption: false,
+        choices: ["keep-original", "keep-latter", "keep-going"],
+        default: "keep-original",
+        desc: "The behavior to have when an error occurs during the migration of an item.",
```
antoinetissier, line 131:
```suggestion
        desc: "The behavior when an error occurs during the migration of an item.",
```

src/migration.types.ts
```diff
@@ -171,3 +171,11 @@ export type MigrationFunction<
     doesReportIncludeStacks: boolean;
   },
 ) => void;
+
+/**
+ * The behavior to have when an error occurs during the migration of an item:
+ * - "keep-original": if anything goes wrong during the migration of an item, the CLI `from-version` of the item is kept.
```
antoinetissier, line 177:
```suggestion
 * - "keep-original": the item is kept untouched.
```

src/migration.types.ts
```diff
@@ -171,3 +171,11 @@ export type MigrationFunction<
...
+ * - "keep-latter": if anything goes wrong during the migration of an item between n-version and n+1-version, the n-version of the item is kept.
```
antoinetissier, line 178:
```suggestion
 * - "keep-last-successful-version": the version of the item obtained after the p-1 first successful steps is kept, and the n-p remaining steps are not applied.
```
"stop" would be a simpler alternative, but I am afraid it could be misread as something global, as opposed to just one item.

Not a strong opinion though if you prefer "stop".

src/migration.types.ts
```diff
@@ -171,3 +171,11 @@ export type MigrationFunction<
...
+ * - "keep-going":  if anything goes wrong during the migration of an item, keep trying to migrate it, even though it's corrupt, until the CLI `to-version`.
```
antoinetissier, line 179:
```suggestion
 * - "keep-going":  the n-p remaining steps are applied on the version of the item obtained after the first p-1 successful steps.
```

We could also go for "skip". What do you think ?

src/migration.types.ts
```diff
@@ -171,3 +171,11 @@ export type MigrationFunction<
     doesReportIncludeStacks: boolean;
   },
 ) => void;
+
+/**
+ * The behavior to have when an error occurs during the migration of an item:
```
antoinetissier, line 176:
```suggestion
 * The behavior when an error occurs during the migration of an item.
 * Assuming that the error occurred at step p out of a total of n:
```

antoinetissier:
> That means if something is wrong with the migration of a widget within a dashboard, the option will only apply to that widget migration and not to the rest of the dashboard state. In such a case, the counters.dashboards.partial variable is incremented.

Widgets within a dashboard won't be affected by the option: it would be like a "keep-going" for them.
The option will only impact errors occuring within `migrateDashboard` but outside of `migrateWidgetsWithinDashboard`.

src/migration.types.ts
eliasinho: Yeah, I'm not a fan of `"stop"`, as you said it's too generic, and could be misleading.
I find `"keep-last-successful-version"` a bit long, what do you think about more simply `"keep-last-successful"`?

src/migration.types.ts
eliasinho: I find `"keep-going"` clearer and actually more accurate. With `"skip"`, I feel like people could understand "skip the migration", as in "don't migrate this item at all", instead of "skip the error". What do you think?

eliasinho:
> Widgets within a dashboard won't be affected by the option: it would be like a "keep-going" for them. The option will only impact errors occuring within `migrateDashboard` but outside of `migrateWidgetsWithinDashboard`.

Why wouldn't they be affected? And why `"keep-going"` would be the default behavior for them?

src/migration.types.ts
antoinetissier: It's an adjective so I would find it a bit weird: in the eyes of a consumer of the CLI I would wonder "last successful what ?"

src/migration.types.ts
antoinetissier: `"keep-going"` sounds good then

antoinetissier:
> > Widgets within a dashboard won't be affected by the option: it would be like a "keep-going" for them. The option will only impact errors occuring within `migrateDashboard` but outside of `migrateWidgetsWithinDashboard`.
> 
> Why wouldn't they be affected? And why `"keep-going"` would be the default behavior for them?

It's the design that we had agreed on, and that is currently in place for 4.3 to 5.0: you don't stop migrating a dashboard when there is an error in one of the widgets.

src/migration.types.ts
eliasinho: Yeah, it's fair enough. Happy to go for `"keep-last-successful-version"` then, I can't think of anything better.

eliasinho:
> > > Widgets within a dashboard won't be affected by the option: it would be like a "keep-going" for them. The option will only impact errors occuring within `migrateDashboard` but outside of `migrateWidgetsWithinDashboard`.
> > 
> > 
> > Why wouldn't they be affected? And why `"keep-going"` would be the default behavior for them?
> 
> It's the design that we had agreed on, and that is currently in place for 4.3 to 5.0: you don't stop migrating a dashboard when there is an error in one of the widgets.

Ah yes, you keep migrating the rest of the dashboard state, that okay.
But what version of the widget which migration errored do you keep? The original one or the last successful one?

eliasinho:
I slightly reworded the `BehaviorOnError` type description, let me know if you disagree!

antoinetissier:
> But what version of the widget which migration errored do you keep? The original one or the last successful one?

It's going to be the last successful one

src/migration.types.ts
```diff
@@ -176,7 +176,10 @@ export type MigrationFunction<
  * The behavior when an error occurs during the migration of an item.
  * Assuming that the error occurred at step p out of a total of n:
  * - "keep-original": the item is kept untouched.
- * - "keep-last-successful-version": the version of the item obtained after the p-1 first successful steps is kept, and the n-p remaining steps are not applied.
- * - "keep-going":  the n-p remaining steps are applied on the version of the item obtained after the first p-1 successful steps.
+ * - "keep-last-successful-version": the item obtained after the first p-1 successful steps is kept, and the n-p remaining steps are not applied.
+ * - "keep-going": the n-p remaining steps are applied to the corrupt item obtained after the p-th error step.
```
antoinetissier, line 180: I would prefer to avoid corrupt: we already use this wording to refer to a `content` file for which there is no match in `strcuture`.
And the fact that an error occurs does not necessarily mean that a file is "corrupt": maybe the migration script just doesn't handle this case, like for a custom widget.

Maybe: "keep trying to apply the n-p remaining steps even though step n failed".

eliasinho: I tried something, let me know what you think!

antoinetissier:
I added some changes to make it clear that the option only has an effect when migrating through several versions in one go.
Also, I removed the alias as this option is pretty niche and I don't think `-e` is very intuitive: I prefer consumers of the CLI to write `--on-error` explicitly and know what they're doing.

src/migration.types.ts
```diff
@@ -174,10 +174,18 @@ export type MigrationFunction<
...
- * - "keep-original": the item is kept untouched.
- * - "keep-last-successful-version": the item obtained after the first p-1 successful steps is kept, and the n-p remaining steps are not applied.
- * - "keep-going": the n-p remaining steps are applied to the item obtained after step p, despite the error.
+ * This has an effect only when migrating through several versions in one go.
+ *
+ * For example, suppose that you're migrating from 5.0 to 5.3.
+ * For each saved item (e.g. a dashboard), three migration steps are applied: 5.0 => 5.1, 5.1 => 5.2, and 5.2 => 5.3.
+ * Each of these three steps might fail.
+ *
+ * More generically, assuming that the error occurred at step p out of a total of n, you can choose one of the following behaviors:
```
eliasinho, line 183: Maybe now that you gave the example just above, it might be worth completing each case description by explaining what would happen if in the example above the fail happened at step 2?

src/migration.types.ts
```diff
@@ -174,10 +174,18 @@ export type MigrationFunction<
...
+ * - "keep-original": keep the original item untouched, as before any migration step.
```
eliasinho, line 184:
```suggestion
* - "keep-original": keep the original item untouched, as before the whole migration.
```
I was just slightly confused and read that as "before any of the migrations steps" at first. Maybe this is is a bit better?

eliasinho (commented):
I like the improvements! Just made a couple suggestions

src/migration.types.ts
antoinetissier: The example was just meant to explain that the option only has an impact if there are n >= 2 migration steps.
I think it's nice to keep the explanation generic after that.

eliasinho: I meant keep the generic explanations, which I agree are nice, but complete each of them with a quick illustration of what they would give with in your example.

eliasinho (commented):
There's still [this comment](https://github.com/activeviam/activeui-migration/pull/69#discussion_r1083070300), but appart from that it looks good.
I'll it up to you to take that last comment into account or not. Either way, I'm happy for it to be merged in that state (I can't approve as I'm the author)!

src/migration.types.ts
antoinetissier: I'm afraid that it would double the size of the description, which is already rather big.
So if you don't mind I would prefer to keep it this way for now, and we can always change it later if people struggle with it.

Completely unrelated, but I think most consumers will just use this package as a CLI and won't fork it, so they'll never see this JSDoc.
I think it's still nice to keep it internally, but the most important is to have this description in the `yargs` CLI.

eliasinho:
The diff shows quite a lot of lines added, but what was really added here is:
- the introduction of `idsOfDashboardsToMigrate`, `idsOfWidgetsToMigrate` and `idsOfFiltersToMigrate`, which role is to keep track of the items that need to keep being migrated during migration `p+1`: successful item migration or `"keep-going"` behavior on error
- using `behaviorOnError` in `getMigrateXyz`
- introduce `originalContentServer` in case `"keep-original"` is chosen to be able to retrieve the original version of an item after `p` migrations
- adapt the same logic when the migration starts from `4.3`

src/getMigrateDashboards.ts
```diff
@@ -61,101 +70,120 @@ export const getMigrateDashboards =
...
     for (const fileId in content.children) {
-      if (!filesAncestry[fileId]) {
-        counters.dashboards.removed++;
-        _addCorruptFileErrorToReport(errorReport, {
-          contentType: "dashboards",
-          fileId,
-        });
-        continue;
-      }
+      if (idsOfDashboardsToMigrate.has(fileId)) {
```
antoinetissier, line 73: If we go this route, we could go `for fileId of idsOfDashboardsToMigrate` directly.

But if I recall the design session correctly, we had agreed on another design, based on `failedDashboardIds`.
I find it a bit more descriptive/intuitive (as it describes the extra piece of information that we need explicitly rather than implicitly; for example it would make it easier to write unit tests), and it should always be smaller than `idsOfDashboardsToMigrate` (which in the best case scenario is always all the ids).

With the same idea, we could even decide not to introduce a new `failedDashboardIds` variable, since we essentially already have the `errorReport` to store that  information.
If we go for that route we'd need to change things a little bit though if we want to keep the functional design of "always keep going when the migration of a widget within a dashboard fails".

If you had decided to go for `idsToMigrate` on purpose (vs `failedIds`), can you explain why you think it's better ?

src/cli/bin.ts
```diff
@@ -132,8 +149,24 @@ yargs
...
+      // `originalContentServer` will be used if the user wants to keep the original version of an item if an error happens during the migration.
       const contentServer: ContentRecord = await fs.readJSON(inputPath);
+      const originalContentServer = _cloneDeep(contentServer);
+
+      const dashboardsContent =
+        contentServer.children?.ui.children?.dashboards.children?.content;
+      const widgetsContent =
+        contentServer.children?.ui.children?.widgets.children?.content;
+      const filtersContent =
+        contentServer.children?.ui.children?.filters.children?.content;
```
antoinetissier, line 164: This doesn't work for `--from-version="4.3"`

src/getMigrateDashboards.ts
```diff
@@ -53,6 +60,8 @@ export const getMigrateDashboards =
   ): void => {
     const { content, structure } =
       contentServer.children?.ui.children?.dashboards.children ?? {};
+    const originalContent =
+      originalContentServer.children?.ui.children?.dashboard.children?.content;
```
antoinetissier, line 64: This doesn't work for `--from-version="4.3"`, and same for widgets/filters.
Instead, the caller could pass `originalContent` directly, and in practice it would be defined once and for all in `bin.ts`.

antoinetissier (commented):
To take something closer to what we had in the design session, I would have imagined something like this (in pseudocode):
```javascript
function getMigrateDashboards(contentServer, behaviorOnError, cb) {
  contentServer.dashboards.forEach((dashboard, id) => {
    if(failedDashboardIds.has(id) && behavior !== "keep-going") {
      return;
    }
    try {
        const updatedDashboard = produce(dashboard, cb)
        contentServer.push(updatedDashboard)
      } catch(error) {
        failedDashboards.add(id);
        if (behavior === "keep-original"){
          contentServer.push(originalContent.children[id]);
        }
      }
  })
}

```

src/cli/bin.ts
```diff
@@ -122,6 +129,25 @@ yargs
...
+      args.option("on-error", {
+        type: "string",
+        demandOption: false,
+        choices: [
+          "keep-original",
+          "keep-last-successful-version",
+          "keep-going",
+        ],
+        default: "keep-original",
+        desc: `The behavior when an error occurs during the migration of an item. This has an effect only when migrating through several versions in one go.
```
eliasinho, line 141: FYI, this what it gives in a terminal:
image

src/getContent.ts
```diff
@@ -0,0 +1,26 @@
+import { ContentRecord, ContentType } from "@activeviam/activeui-sdk-5.0";
+
+/**
+ * Returns the `contentType` content from the `contentServer` of version `fromVersion`.
+ */
```
antoinetissier, line 5:
```suggestion
/**
 * Returns the content of type `contentType` created with ActiveUI version `fromVersion` on `contentServer`.
 */
```

src/getMigrateDashboards.ts
```diff
@@ -54,13 +60,24 @@ export const getMigrateDashboards =
...
+    if (!content?.children || !structure?.children || !originalContent) {
       return;
     }
 
     const filesAncestry = _getFilesAncestry(structure);
 
     for (const fileId in content.children) {
+      if (
+        errorReport.dashboards?.[fileId] &&
+        !isDashboardErrorReport(errorReport.dashboards?.[fileId]) &&
```
antoinetissier, line 72: Not your fault because the type was already like this, but I find the name confusing, and would go for `isWidgetWithinDashboardErrorReport` instead.

src/getMigrateDashboards.ts
```diff
@@ -54,13 +60,24 @@ export const getMigrateDashboards =
...
+        behaviorOnError !== "keep-going"
+      ) {
+        // The migration of this dashboard failed at a previous step.
+        // It is is not a partial failure (i.e. an error on the migration of one of its widgets).
+        // The behavior on error is not to keep going, hence the migration of this dashboard should not go further.
```
antoinetissier, line 77: I would remove these lines as I think the rest speaks for itself (or will with the renaming suggestion above).
```suggestion
```

src/getMigrateDashboards.ts
```diff
@@ -156,6 +173,12 @@ export const getMigrateDashboards =
           fileId,
           name,
         });
+
+        // If the behavior is "keep-last-successful-version" or "keep-going", the dashboard is kept as it is.
```
antoinetissier, line 177: I think the code speaks for itself here too, as the variables are nicely named.
```suggestion
```

antoinetissier (commented):
In the current state of the PR the errors thrown during the migration from 4.3 to 5.0 do not respect `BehaviorOnError`.
For the other migration steps, the remarks on `getMigrateDashboards` apply to the other content types

src/getMigrateDashboards.ts
eliasinho: Hmm, the parameter is not a widget, so I find this naming confusing.
What about `isWithinDashboardErrorReport`?

antoinetissier: Semantically speaking `WithinDashboard` alone also applies to dashboard filters, page filters, and well, everything within the dashboard.
So we need the extra precision of `isWidgetWithinDashboard`, which also makes it in line with `migrateWidgetsWithinDashboard`.

antoinetissier:
@eliasinho I pushed a commit, can you review it ?
It looks good to me in this state.

antoinetissier (approved):

src/isPartialDashboardErrorReport.ts
```diff
@@ -5,7 +5,7 @@ import {
 
 /**
  * Asserts whether `fileErrorReport` is a {@link PartialDashboardErrorReport}.
- * Partial dashboard errors are reported when the migration of a dashboard is successful was a whole, but the migration of some widgets within it failed.
+ * Partial dashboard errors are reported when the migration of a dashboard is successful as a whole, but the migration of some widgets within it failed.
```
antoinetissier, line 8:
```suggestion
* Partial dashboard errors are reported when the migration of a dashboard was successful as a whole, but the migration of some widgets within it failed.
```

PR #72, opened 2023-01-13
UI-7645 - Refactor types for extra migration callback args

antoinetissier:
Adding an extra `Options` generic for the migration callbacks, representing the extra options that one specific migration function might have.
For example `namesOfCalculatedMeasuresToMigrate` is necessary for the dashboard/widget migration callback of 5.0 to 5.1, but it most likely won't be for 5.1 to 5.2.
Doing it this way allows to assign `MigrateXyzCallback` to some variables while still keep the factorization for the other common args (e.g. `dataModels`)

### DIFF
--- a/src/5.0_to_5.1/migrate_5.0_to_5.1.ts
+++ b/src/5.0_to_5.1/migrate_5.0_to_5.1.ts
@@ -1,12 +1,12 @@
 import {
-  deserializeWidgetState,
-  deserializeDashboardState,
   parse,
+  deserializeDashboardState,
+  deserializeWidgetState,
 } from "@activeviam/activeui-sdk-5.0";
 import {
+  stringify,
   serializeWidgetState,
   serializeDashboardState,
-  stringify,
 } from "@activeviam/activeui-sdk-5.1";
 import { MigrationFunction } from "../migration.types";
 import { migrateDashboard } from "./migrateDashboard";
@@ -21,6 +21,7 @@ export const migrate_50_to_51: MigrationFunction = (
     migrateDashboards,
     migrateSavedWidgets,
     migrateSavedFilters,
+    dataModels,
     errorReport,
     counters,
     doesReportIncludeStacks,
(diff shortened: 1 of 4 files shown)


src/migration.types.ts
```diff
@@ -85,13 +85,13 @@ export interface ErrorReport {
...
+  ToDashboardState,
+  Options extends Record<string, unknown> = Record<string, unknown>,
+> = (
   dashboardState: FromDashboardState,
-  {
-    dataModels,
-    keysOfWidgetPluginsToRemove,
-    onErrorWhileMigratingWidget,
-  }: {
+  args: {
```
eliasinho, line 94:
```suggestion
  options: {
```
It's not a huge deal, but I'm not a fan of `args`: `dashboardState` is also an argument, and the main one actually. To me those are more "options". However the type is called `Options` already.
We could:
- rename `args` in `options` and the type in `ExtraOptions`
- rename `args` in `baseOptions` and keep the type as `Options`
- rename `args` in `baseOptions` and the type in `ExtraOptions`

I prefer the first option, but if you disagree feel free to go for another option (no pun intended lmao), again I don't think it's a big deal.

src/migration.types.ts
```diff
@@ -85,13 +85,13 @@ export interface ErrorReport {
  * A function that can be called to migrate a dashboard from one version to another.
  * It is called within {@link produce | https://immerjs.github.io/immer/produce}, so it can safely mutate its input `dashboardState`.
  */
-export type MigrateDashboardCallback<FromDashboardState, ToDashboardState> = (
+export type MigrateDashboardCallback<
+  FromDashboardState,
+  ToDashboardState,
+  Options extends Record<string, unknown> = Record<string, unknown>,
```
eliasinho, line 91: Is `unkown` better than `any`?

src/5.0_to_5.1/migrate_5.0_to_5.1.ts
```diff
@@ -35,22 +36,21 @@ export const migrate_50_to_51: MigrationFunction = (
     deserializeDashboardState,
     (
       dashboardState,
-      { dataModels, keysOfWidgetPluginsToRemove, onErrorWhileMigratingWidget },
```
eliasinho, line 38: Do we still need `dataModels` in the type of the callbacks?

src/migration.types.ts
antoinetissier: Yes, `Record<string, unknown>` is recommended type to represent "any object"

src/5.0_to_5.1/migrate_5.0_to_5.1.ts
antoinetissier: Yes we do, otherwise you could not consume it with just the `MigrateXyzCallback` type.
But I'm not sure I understood your question

src/migration.types.ts
antoinetissier: What I don't like in  "options" is that `dataModels` is not optional for example.
Whereas things like `namesOfCalculatedMeasuresToMigrate` is an opt-in of 5.0 of 5.1, but it's not there for every migration step.
That's why I went for `args` for the whole object but `Options` for those extra opt-in things.

eliasinho: Fair enough, it's a good point. Let's go with `args`!

eliasinho (approved):
Looks good! ✔

PR #73, opened 2023-01-18
UI-7645 - Bump 5.1 snapshot version to align on the new filter types

antoinetissier:
<h2>Overview</h2>

On the latest version of the SDK we're going from storing the MDX of filters to storing `Filter`, i.e. an interpretation of the MDX, easier to use and with more info.
This PR aligns the migration scripts on it.

<h2>Problem 1</h2>

In this new version of the SDK, packages are not shipped with a CJS bundle anymore.
Therefore, to make the tests work when run with Jest (which expects CJS by default), they need to be transpiled.
This is what `activeui-sdk-scripts` does, but it is no longer published (its package.json now has `private: true`).
So we need to use Jest directly instead (and it is simpler), with a specific config.
The latest version being Jest 29, the snapshot format has changed.
You can check that only the format is indeed changed and not the content.

<h2>Problem 2</h2>

This dependency upgrade reveals a rather complex issue of peer dependencies, which is why there are more changes to this PR than changing the sdk version in `package.json`.
Here is an attempt to summarize it.

<h3>Context before this PR</h3>

`activeui-sdk` has peer dependencies such as `antd` and `react`.
Peer dependencies are NOT installed by `yarn`.
So building the package or running tests would fail because of dependency chains such as this example one: `import {serializeDashboardState} from @activeviam/dashboard-base` => ... => `import {getAntdTheme} @activeviam/theme` => `import {theme} from antd` (this is an example with the more recent activeui-sdk-5.1 version; but the idea is the same).
So we had to install these dependencies ourselves "manually", by adding them as `devDependencies` of `activeui-migration`: see https://github.com/activeviam/activeui-migration/commit/fc292233d9e66be8bfb237e6df07bc91564639b4

<h3>Problem in this PR</h3>

`activeui-sdk-5.0` dependency transitively depends on `antd-4` and `react-17` but the latest `activeui-sdk-5.1` depends on `antd-5` and `react-18`.
We need to import both sdk versions (e.g. to migrate from 5.0 to 5.1 we need `deserializeDashboardState` from 5.0 and `serializeDashboardSTate` from 5.1).
=> For the reasons explained above, we'd need to install both `antd-4` and `antd-5` and `react-17` and `react-18`, which is not possible by just adding them to our `devDependencies`.

<h3>Solution implemented in this PR</h3>

These peer dependencies are UI/browser related.
We don't need them to run the `activeui-migration` CLI and they do not make sense in a Node environment.
So instead of installing them + bundling them with Webpack + going through the trouble of transpiling them when running tests, it is better to mock them.
This makes the `node_modules` lighter, the `activeui-migration` package lighter, and should make the tests a bit faster too.
So as you can see in the PR changes, these dependencies are removed from the `package.json`, and mocked in the `__mocked__` folder instead, which in particular is used by Jest.

<h3>Alternative solution, to potentially implement after 5.1</h3>
Conceptually, not installing and mocking dependencies that are not needed is nice.
However, I don't think this scales very well: in particular we're doing it for `react` and `antd` because as peer dependencies they didn't get installed, but there are many other browser/UI related libraries that we don't mock => so in that sense there is an inconsistency.
It's also not easy to know precisely what to mock (although the top level API of those packages is a good start), so it can be hard to maintain.

So if this ends up being a pain in the future, we could roll back on the concept, and instead find a way to install those peer dependencies without conflict.
We could do this by (to verify):
- using a dedicated package https://yarnpkg.com/package/install-peers 
- using `npm > 7`
- using `yarn 2` https://yarnpkg.com/features/pnp

=> less good perf wise and bigger changes, but should be much simpler to maintain
Created UI-8152 for it.


yarn.lock
```diff
@@ -3448,22 +3412,20 @@
     "@activeviam/icons" "5.0.15"
     "@activeviam/theme" "5.0.15"
     "@emotion/core" "^10.0.27"
-    antd "4.12.2"
```
antoinetissier, line 3451: ⚠️ This was a manual change: in this package antd is mistakenly a dependency (see https://github.com/activeviam/activeui/blob/205ad3f27e23a08c6e6c7cc647fd9b0e53e90587/packages/plugin-widget-text-editor/package.json#L27) instead of being a peer dependency.

babel.config.js
```diff
@@ -0,0 +1,18 @@
+// Used by Jest to transpile files before running test.
+// The `transformIgnorePatterns` propery in the Jest config controls which files the transformations are applied to.
```
messaooudi, line 2:
```suggestion
// The `transformIgnorePatterns` property in the Jest config controls which files the transformations are applied to.
```

jest.config.js
```diff
@@ -1,31 +1,31 @@
...
-  "dnd-core",
-];
+const esmPackages = ["@activeviam/*", "lodash-es", "monaco-editor"];
 
-module.exports = {
-  ...jestConfig,
-  reporters: ["default"],
+// Note that in addition to the config below, there are several mocked modules under the __mocks__ folder.
+// They correspond to modules that are UI/browser-related and are transitively pulled by @activeviam dependencies, but are not necessary for `activeui-migration` to work in a Node environment.
+const config = {
```
messaooudi, line 9: Can we use the config exported by `@activeviam/jest-config-activeui` package ?

babel.config.js
```diff
@@ -0,0 +1,18 @@
+// Used by Jest to transpile files before running test.
```
messaooudi, line 1: This is needed because you dropped the use of `--experimental-vm-modules` right? If yes, why ?

jest.config.js
antoinetissier: No as this package is private.

babel.config.js
antoinetissier: My understanding is that you have two choices:
- either migrate the package you want to test to ESM, and use `--experimental-vm-modules`
- or make sure that the ESM packages you depend on are transpiled to CJS

I chose the second option because it seems more atomic: migrating `activeui-migration` to pure ESM seems more out of scope of this PR.
This is also in line with what we did in the ActiveUI monorepo: first we made sure that `Jest` could still be ran in packages that they were not fully ESM yet (most of our packages), and then progressively we migreate each package to ESM, separately.

messaooudi (approved):
Looks good

PR #74, opened 2023-01-24
UI-7645 - Filters migration

antoinetissier:
Migrates filters from `MdxExpression` in 5.0 to `Filter` in 5.1.
Saved filters had an extra wrapper: `{mdx: MdxExpression}`, but in 5.1 saved filters and filters within widgets/dashboards are exactly the same.

### DIFF
--- a/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
+++ b/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
@@ -23,7 +23,7 @@ export function createFilterWithFirstMatchingCube(
   }
 
   throw new Error(
-    `The following MDX does not represent a filter: ${stringify(mdx, {
+    `No cube contains the hierarchies expressed in the following filter MDX: ${stringify(mdx, {
       indent: true,
     })}`,
   );
--- a/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
+++ b/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
@@ -15,7 +15,6 @@ export function createFilterWithFirstMatchingCube(
 ): Filter {
   for (const cube of cubes) {
     try {
-      console.log(cube)
       return createFilter(mdx, cube);
     } catch (e) {
       // The filter creation might be successful with another cube.
--- a/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
+++ b/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
@@ -7,7 +7,7 @@ import {
 } from "@activeviam/activeui-sdk-5.1";
 
 /**
- * Returns a {@link Filter} based on the input `mdx`, with the first matching cube.
+ * Returns a {@link Filter} based on `mdx`, with the first cube containing the hierarchies expressed in the MDX expression.
  */
 export function createFilterWithFirstMatchingCube(
   mdx: MdxExpression,
--- a/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
+++ b/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
@@ -7,7 +7,7 @@ import {
 } from "@activeviam/activeui-sdk-5.1";
 
 /**
- * Returns a {@link Filter} based on `mdx`, with the first cube containing the hierarchies expressed in the MDX expression.
+ * Returns a {@link Filter} based on `mdx`, with the first cube containing the hierarchies expressed in `mdx`.
  */
 export function createFilterWithFirstMatchingCube(
   mdx: MdxExpression,
--- a/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
+++ b/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
@@ -15,6 +15,7 @@ export function createFilterWithFirstMatchingCube(
 ): Filter {
   for (const cube of cubes) {
     try {
+      console.log(cube)
       return createFilter(mdx, cube);
     } catch (e) {
       // The filter creation might be successful with another cube.
@@ -23,6 +24,8 @@ export function createFilterWithFirstMatchingCube(
   }
 
   throw new Error(
-    `The following MDX does not represent a filter: ${stringify(mdx)}`,
+    `The following MDX does not represent a filter: ${stringify(mdx, {
+      indent: true,
+    })}`,
   );
 }
--- a/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
+++ b/src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
@@ -23,8 +23,11 @@ export function createFilterWithFirstMatchingCube(
   }
 
   throw new Error(
-    `No cube contains the hierarchies expressed in the following filter MDX: ${stringify(mdx, {
-      indent: true,
-    })}`,
+    `No cube contains the hierarchies expressed in the following filter MDX: ${stringify(
+      mdx,
+      {
+        indent: true,
+      },
+    )}`,
   );
 }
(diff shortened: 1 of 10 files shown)


src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
```diff
@@ -0,0 +1,30 @@
+import { stringify } from "@activeviam/activeui-sdk-5.0";
+import {
+  createFilter,
+  MdxExpression,
+  Filter,
+  Cube,
+} from "@activeviam/activeui-sdk-5.1";
+
+/**
+ * Returns a {@link Filter} based on the input `mdx`, with the first matching cube.
```
eliasinho, line 10:
```suggestion
 * Returns a {@link Filter} based on `mdx`, with the first matching cube.
```
Also, can we be a bit more specific on what "matching" means here?

src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
```diff
@@ -0,0 +1,30 @@
...
+    try {
+      return createFilter(mdx, cube);
+    } catch (e) {
+      // The filter creation might be successful with another cube.
+      // Continue.
+    }
+  }
+
+  throw new Error(
+    `The following MDX does not represent a filter: ${stringify(mdx, {
```
eliasinho, line 26: Is that accurate?

If I understand correctly, getting here means that `createFilter` threw for every cube. And `createFilter` can indeed throw if `mdx` doesn't represent a filter, but also if the coordinates specified in `mdx` don't "match" what is specified `cube`. Right?

src/5.0_to_5.1/migrateFilters.test.ts
```diff
@@ -1,12 +1,15 @@
...
 describe("migrateFilters", () => {
   it("wraps the MDX representing each filter", () => {
     const euroAsString = "[Currency].[Currency].[ALL].[AllMember].[EUR]";
     const euro = parse(euroAsString);
     const filters = [euro];
-    migrateFilters(filters);
+    migrateFilters(filters, { dataModels });
     expect(filters).toHaveLength(1);
     // @ts-expect-error The point of the migration is to mutate the filters from `mdx` to `{mdx}`.
     expect(stringify(filters[0].mdx)).toBe(euroAsString);
```
eliasinho: I'm not sure I understand how that works: unless I'm missing something, `Filters` doesn't have an `mdx` attribute

src/5.0_to_5.1/migrateFilters.test.ts
```diff
@@ -1,12 +1,15 @@
 import { parse, stringify } from "@activeviam/activeui-sdk-5.0";
+import { sandboxDataModel } from "@activeviam/data-model-5.1/dist/__test_resources__";
 import { migrateFilters } from "./migrateFilters";
 
+const dataModels = { sandbox: sandboxDataModel };
+
 describe("migrateFilters", () => {
   it("wraps the MDX representing each filter", () => {
```
eliasinho: Would be cool to add two other tests:
- one with `cubeName` explicitly passed to `migrateFilters`, to cover that case
- one where no cube is matching the `mdx`, to illustrate the logic of `createFilterWithFirstMatchingCube` and in which case it would throw

src/5.0_to_5.1/migrateFilters.test.ts
eliasinho: Never mind, I just saw your last commit

src/5.0_to_5.1/migrateWidget.ts
```diff
@@ -25,6 +29,16 @@ export const migrateWidget: MigrateWidgetCallback<
     namesOfCalculatedMeasuresToMigrate,
     measureToCubeMapping,
   });
-  migrateFilters(widgetState.filters);
+
+  if (isWidgetWithQueryState(widgetState)) {
```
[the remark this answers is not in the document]
antoinetissier: It's a valid use case when you're building a dashboard: start from an empty pivot table, add filters => at this stage `widgetState.filters` is defined but `widgetState.query.mdx` isn't, and then only add stuff on rows => here `widgetState.query.mdx` gets populated.
Honestly I'd be surprised if people had saved empty widgets, but still we should cater fo rit and it's not more expensive for us if we want to respect the types anyway.

src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
antoinetissier: During the migration the client is expected to pass all of the cubes that they're using, through `servers.json`.
If there is no cube for which this filter can be created, then from their perspective it doesn't represent a filter imo.

Would you phrase things differently ?

Also, an alternative would be to transform it as a custom filter instead. Potentially with a message like "error while migrating filter {id} with mdx {mdx}" in the report.

What do you think ?

src/5.0_to_5.1/migrateFilters.test.ts
antoinetissier: You're right that's outdated, I updated the test.

src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
eliasinho: > Would you phrase things differently ?

I guess I'd rather have a message closer to what's in the description:
```suggestion
    `No cube contains the hierarchies expressed in the following filter MDX: ${stringify(mdx, {
```

> an alternative would be to transform it as a custom filter instead

What is a custom filter exactly?

> Potentially with a message like "error while migrating filter {id} with mdx {mdx}" in the report.

Actually I think we should do that either way. If we throw here the error will be caught higher up, and the error will be added to the error report. So we might as well as be specific as possible and specify both the filter id and the mdx, right?

src/5.0_to_5.1/migrateFilters.test.ts
```diff
@@ -1,14 +1,83 @@
...
+            ],
+          ],
+          "type": "members",
+        },
+      ]
+    `);
+  });
+
+  it("throws if there is no cube containing the hierarchies expressed in Filter", () => {
+    const squirtle = "[Pokemon].[Water].[ALL].[AllMember].[Squirtle]";
```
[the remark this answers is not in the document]
antoinetissier: Proper thicc

src/5.0_to_5.1/migrateFilters.ts
```diff
@@ -1,13 +1,38 @@
...
+  DataModel,
+} from "@activeviam/activeui-sdk-5.1";
+import { Mdx } from "@activeviam/activeui-sdk-5.0";
+import { getAllCubes } from "./getAllCubes";
+import { createFilterWithFirstMatchingCube } from "./createFilterWithFirstMatchingCube";
 
 /**
- * Wraps each filter in `filters` under an `mdx` property in an object.
- * Indeed, in 5.1 filters can have an extra `isPinned` property.
+ * In 5.0, a filter in a widget/dashboard state is just an {@link MdxExpression}.
```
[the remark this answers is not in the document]
antoinetissier: Then eslint will complain that it's imported but never used.
And if we cannot make it poka-yoke we're sure that this will not be respected in a bunch of places, so I prefer to keep it simple for now.

src/5.0_to_5.1/createFilterWithFirstMatchingCube.ts
antoinetissier: Actually this function is low level and is agnostic of the `id`, and there is not even necessarily an `id` if it's just a filter from a dashboard or widget state.
But if it's a saved filter being migrated then indeed the error will be caught higher up and reported with the file id.

src/5.0_to_5.1/migrateFilters.ts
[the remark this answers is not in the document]
eliasinho: Are you sure? I just quickly checked with a dummy file and it seems fine:
image
image

eliasinho (approved):
Looks great! There's still [this comment](https://github.com/activeviam/activeui-migration/pull/74#discussion_r1087631599) regarding the link to the types in the function description.

But I'll approve still to unblock this PR.

antoinetissier:
In `activeui-migration` I had the error on my IDE. Maybe it's just a config thing, but if you don't mind I'll merge the PR like this to not lose time on it.

PR #75, opened 2023-01-26
UI-8110 - Remove the non-nullable cast for AMetaData.name

mailys-ds:
Close [UI-8110](https://support.activeviam.com/jira/browse/UI-8110).


antoinetissier (approved):
💯

PR #76, opened 2023-01-26
UI-8168 - Add fail version to the error report

eliasinho:
## Description

This PR adds the fail version to the error report.
If the migration of an item failed from version p to version p+1, the p+1 version is specified as fail version.


antoinetissier (commented):
I find the current state a bit confusing: one could read `failVersion` as "the version of the item which could not be migrated", even though that's not what you meant.
IMO it would be clearer to have a `step` property whose value would be `"4.3 to 5.0"` for example.

And instead of hardcoding it, it would be nice to make it generic by passing `from` and `to` to the `getMigrateXyz` functions.

antoinetissier:
Looks good ! ✅ What would be nice is to check that if we have `keep-latest-version` as behavior on error, we potentially get errors from different steps in the report.
But it's not a priority and we can always improve on it later, let's merge like this 🚀

PR #77, opened 2023-01-27
UI-7065 - Migrate widget plugin key of combo charts

antoinetissier:
UI-7065
https://github.com/activeviam/activeui/pull/3243


eliasinho (approved):
Nice!

PR #79, opened 2023-01-31
UI-8181 - Fix wrong counter being incremented

eliasinho:
## Description

`counters.dashboards.removed` was incremented instead of `counters.widgets.removed` and `counters.filters.removed`. This PR fixes that.


antoinetissier (approved):
Good catch !

PR #80, opened 2023-01-31
UI-8182 - Increment widgets.removed counter when widget key in keysOfWidgetPluginsToRemove

eliasinho:
## Description

The `widgets.removed` counter is not incremented when the key of a widget within a dashboard is specified in the `--remove-widgets`. This PR fixes that.


antoinetissier (approved):
Nice 👍

PR #82, opened 2023-01-31
UI-8176 - Make property `version` of `WidgetMetaData` mandatory

mailys-ds:
Close UI-8176.

### DIFF
--- a/src/5.0_to_5.1/__test_resources__/contentServer.ts
+++ b/src/5.0_to_5.1/__test_resources__/contentServer.ts
@@ -618,8 +618,7 @@ export const contentServer: ContentRecord = {
                   children: {
                     "663_metadata": {
                       entry: {
-                        content:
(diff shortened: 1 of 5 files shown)


antoinetissier (commented):
In the current state of the PR, the `version` attribute is added on the fly  when reading the metadata, but it doesn't migrate anything.

Instead, each saved widget metadata record itself should be migrated.
`_getMetaData` doesn't need to be touched, and instead this can be done in `migrate_50_to_51`.

antoinetissier (commented):
Let's remove the `console.log`s but otherwise it's the right idea 👍 
It'd be nice to add a unit test, and more importantly to test the CLI on a content server having saved widgets with/without a `version` attribute in their metadata.

eliasinho:
I tried the with CLI and the migration works well!

antoinetissier (approved):
Looks good 👍

PR #83, opened 2023-01-31
UI-8188 - Remove corrupt files from content server

eliasinho:
## Description

A corrupted file is a file which content is saved, but is not present in the structure. At the moment, during the migration we spot them, account for them in the error report and counters, but we don't actually remove them.


antoinetissier (approved):
Good catch ! 💯

PR #84, opened 2023-01-31
UI-6616 - Prepare 5.1.0 release

antoinetissier:
Bump the `@activeviam` dependencies after https://github.com/activeviam/activeui/commit/12b0370df16bc50394c297f9e4d9d1d49bf451d6

**Node related issues after https://github.com/activeviam/activeui/commit/12b0370df16bc50394c297f9e4d9d1d49bf451d6**
This commit forces the use of Node >= 18.12.1
We need to align on it in the CLI, and in the CircleCI config.
But this leads to build errors.

[This](https://stackoverflow.com/a/73027407/5036255)  suggests that using `webpack` >= 5.64.0 and using `hashFunction.output = "xxhash64"` should solve the issue.
However, it doesn't unfortunately.
I am still keeping those changes in the PR since it doesn't hurt to use a more recent version of webpack and since `xxhash64` is going to be the default in future versions (see the [doc](https://webpack.js.org/configuration/output/#outputhashfunction)).

As suggested [here](https://stackoverflow.com/a/69699772/5036255), I had to resort to using `NODE_OPTIONS=--openssl-legacy-provider webpack`. This is not ideal but I think it's harmless for a CLI like this: we happen to bundle it with Webpack which thinks it's going to be used as an app in a browser environment, but it won't.

And anyway, it would be nice to stop using Webpack just to mock some browser globals and UI-related dependencies such as React: see UI-8152  

**antd related issues after https://github.com/activeviam/activeui/commit/db03db541c4f785c97b34c56358eeac0e341da33**

We import directly from "antd" now in the 5.1 SDK.
But we still import from submodules like "antd/lib/button" from the 5.0 SDK.
So both need to be mocked.
Again, we should be able to get rid of all of this when we do UI-8152

**misc**
I took the opportunity to use `@swc/jest` instead of `babel-jest` as we now recommend in the migration guide, and because, quoting it (and also having tested it), it is "blazingly fast" ⚡


tibdex (approved):
The diff looks good to me.

Nice job finding solutions for the multiple issues so quickly (and late 🌚)!

webpack.config.js
```diff
@@ -67,6 +68,11 @@ module.exports = {
       react: path.resolve(__dirname, "__mocks__", "react"),
       "react-dom": false,
       "react-intl": false,
+      // Mocks exact imports from antd, such as "import { Button } from antd".
+      // This is required because of the activeui-sdk-5.1 dependency.
+      antd$: path.resolve(__dirname, "__mocks__", "antd"),
+      // Mocks import from antd submodules, such as "import Button from antd/lib/button"
+      // This is required because of the activeui-sdk-5.0 dependency.
```
eliasinho, line 75:
```suggestion
      // Mocks exact imports from "antd", such as `import { Button } from "antd"`.
      // This is required because of the "@activeviam/activeui-sdk-5.1" dependency.
      antd$: path.resolve(__dirname, "__mocks__", "antd"),
      // Mocks import from "antd" submodules, such as `import Button from "antd/lib/button"`.
      // This is required because of the "@activeviam/activeui-sdk-5.0" dependency.
```

eliasinho (approved):
Looks good! Just one nitpick suggestion on the comment, but I'm still approving to enable the merge ✅✅

PR #86, opened 2023-02-02
UI-8195 - Migrate 5.0 settings to permissions

eliasinho:
## Description
This PR adds a migration script to move the following attributes from being settings to become permissions:
- `userFilters.areEnabled` => `canUseUserFilters`
- `userQueryContext.isEnabled` => `canUseUserQueryContext`
- `calculatedMeasures.areEnabled` => `canUseCalculatedMeasures`

If a replacing permission is already be in place, its value shouldn't be overridden.


src/5.0_to_5.1/migrateSettings.ts
```diff
@@ -0,0 +1,73 @@
...
+const settingsToPermissionsMapping = {
+  "userFilters.areEnabled": "canUseUserFilters",
+  "userQueryContext.isEnabled": "canUseUserQueryContext",
+  "calculatedMeasures.areEnabled": "canUseCalculatedMeasures",
+};
+
+/**
+ * Moves some attributes from `settings` to `permissions` and updates their name.
+ * Only moves an attribute if the corresponding permission doesn't exist yet.
+ * If there is no `permissions`, deletes the `settings` attribute that were supposed to be moved.
```
eliasinho, line 12: In this case, the migration fails: the settings were not moved anywhere. I'm not sure we can just create a `permissions` folder out of the blue. .
Should we raise an error message to tell the users the migration failed and that they should manually create a permissions folder? And not delete any setting I guess (I did it for now to have a default behavior until we reach a decision).

tibdex: I think just deleting the setting is dangerous. What's the drawback of creating the `permissions` folder for them? If it's a big deal then I would go for the next best & safe thing: raising an error asking the user to create the folder themself and then rerun the migration script.

src/5.0_to_5.1/migrateSettings.ts
```diff
@@ -0,0 +1,73 @@
...
+    permissions.entry.content = JSON.stringify(permissionsContent);
+  }
+};
+
+/**
+ * Transforms some settings attributes into permissions attributes.
+ * Applies the logic to the organization settings and permissions, as well as to each user's settings and permissions.
+ * Mutates `contentServer`.
+ */
+export const migrateSettings = (contentServer: ContentRecord): void => {
```
tibdex, line 53: Would `uiRootRecord` be a better name than `contentServer`? If yes, please also change it in `migrateSavedCalculatedMeasures` and all the other places.

tibdex (commented):
The code and test look good. Just some behavior remark.

src/5.0_to_5.1/migrateSettings.ts
eliasinho: We call it `contentServer` because we actually pass the object describing the whole content server, and not only its `ui` part. Why were you thinking of `uiRootRecord` as a better name?
Note that we're using the term `contentServer` pretty much everywhere at the moment, so we're taking about the same thing, the change would be at a bigger scale than just the 5.0 --> 5.1 migration.

tibdex: Yeah my bad I read too fast. I would suggest the `rootRecord` name then instead of `contentServer` since the variable does not offer interaction with a content server: it's the record stored at its root.

Let's not do the change now since it's out of scope of this PR.

src/5.0_to_5.1/migrateSettings.ts
eliasinho: Actually you're right, I think we can just create it manually we some default attributes, I don't think it would cause any drawbacks 🤔
Let me give that a try

src/5.0_to_5.1/migrateSettings.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    if (settings) {
+      if (!permissions) {
+        // Creates the "permissions" file before doing the move.
+        users[userName].children.permissions = {
+          entry: {
+            content: "{}",
+            isDirectory: false,
+            owners: ["ROLE_CS_ROOT"],
+            readers: ["ROLE_USER"],
```
tibdex, line 81: Shouldn't we use:

```suggestion
            readers: [userName],
```

so that it's only available to this user?

tibdex (approved):
One last comment but this looks good 👍

src/5.0_to_5.1/migrateSettings.ts
eliasinho: Good point!

src/5.0_to_5.1/migrateSettings.ts
```diff
@@ -0,0 +1,90 @@
...
+        entry: {
+          content: "{}",
+          isDirectory: false,
+          owners: ["ROLE_CS_ROOT"],
+          readers: ["ROLE_USER"],
+        },
+      };
+      organizationPermissions =
+        contentServer.children!.ui.children!["organization_permissions"];
+    }
```
antoinetissier, line 61: Instead of doing this here and using `let`, we could use a ternary expression to create the file if it doesn't exist yet.
It's good even if it ends up not being used: it shows users of admin-ui that it can be used, and is alrady set with the right permissions.

src/5.0_to_5.1/migrateSettings.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    moveSettingsToPermissions(organizationSettings, organizationPermissions);
+  }
+
+  const users = contentServer.children?.ui.children?.users.children || {};
+
+  for (const userName in users) {
```
antoinetissier, line 68: Nitpick but here I would just do an early return if `users` is not defined.

src/5.0_to_5.1/migrateSettings.ts
```diff
@@ -0,0 +1,90 @@
...
+    const settings = users[userName].children?.settings;
+    let permissions = users[userName].children?.permissions;
+
+    if (settings) {
+      if (!permissions) {
```
antoinetissier, line 73: Sme here: we could create the file if it doesn't exist yet, independently of whether there are settings or not.

src/5.0_to_5.1/migrateSettings.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      {
+        "canManageContent": true,
+        "canShare": true,
+        "canUseUserQueryContext": false,
+      }
+    `);
+
+    migrateSettings(contentServer);
+
+    // "calculatedMeasures.areEnabled", "userFilters.areEnabled" and "userQueryContext.isEnabled" are moved from the settings to become permissions.
```
antoinetissier, line 65:
```suggestion
    // "calculatedMeasures.areEnabled", "userFilters.areEnabled" and "userQueryContext.isEnabled" are moved from settings to permissions, with different keys.
```

src/5.0_to_5.1/migrateSettings.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+describe("migrateSettings", () => {
+  let contentServer: ContentRecord;
+
+  beforeEach(() => {
+    contentServer = _cloneDeep(testContentServer);
+  });
+
+  it("doesn't do anything if there is no settings", () => {
+    const contentServerClone = _cloneDeep(contentServer);
```
antoinetissier, line 14: Why clone the content server a second time ?
We already have `testContentServer` and `contentServer`

antoinetissier (commented):
A fes suggestions but it looks good overall !

antoinetissier (approved):
Looks good to me like this !

PR #87, opened 2023-02-03
UI-8196 - Mutate user activity from 5.0 to 5.1

dustinstender:
Now user activity (`userFilters` and `userQueryContext`) will be mutated to usable data for 5.1.


src/5.0_to_5.1/migrateUserActivity.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (users?.children === undefined) {
+    return;
+  }
+  Object.values(users.children).forEach((user) => {
+    if (user.children?.activity) {
+      const userActivity = JSON.parse(user.children?.activity.entry.content);
+      const userFilters = userActivity.userFilters.map(parse);
+      migrateFilters(userFilters, { dataModels });
+      const stringifiedUserFilters = userFilters.map(serializeFilter);
+      userActivity.userFilters = stringifiedUserFilters;
```
eliasinho, line 25:
```suggestion
      userActivity.userFilters = userFilters.map(serializeFilter);
```
Makes you save a line and a long variable name here

src/5.0_to_5.1/migrateUserActivity.ts
```diff
@@ -0,0 +1,31 @@
...
+      migrateContextValues(userActivity.userQueryContext);
+      const stringifiedUserActivity = JSON.stringify(userActivity);
+      user.children.activity.entry.content = stringifiedUserActivity;
```
eliasinho, line 28:
```suggestion
      activity.entry.content =  JSON.stringify(userActivityContent);
```
Same here, saving one line.
(Also using the naming I suggested previously)

src/5.0_to_5.1/migrateUserActivity.ts
```diff
@@ -0,0 +1,31 @@
+import { ContentRecord, parse } from "@activeviam/activeui-sdk-5.0";
+import { serializeFilter } from "@activeviam/dashboard-base-5.1";
+import { DataModel } from "@activeviam/data-model-5.1";
+import { migrateContextValues } from "./migrateContextValues";
+import { migrateFilters } from "./migrateFilters";
+
+/**
+ * Mutates 5.0 `userFilters` and `userQueryContext` into usable data in 5.1.
```
eliasinho, line 8:
```suggestion
 * Migrates the 5.0 user activity into one usable in 5.1.
 * In particular, migrates the user filters and the user query context.
 * Mutates `contentServer`.
```

src/5.0_to_5.1/migrateUserActivity.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  dualCatalog: dualCatalogDataModel,
+};
+
+describe("migrateUserActivity", () => {
+  it("migrates 5.0 `userFilters` and `userQueryContext` to usable data in 5.1", () => {
+    migrateUserActivity(contentServer, dataModels);
+    expect(
+      contentServer.children?.ui.children?.users.children?.user2.children
+        ?.activity.entry.content,
+    ).toMatchInlineSnapshot(
```
eliasinho, line 19: I think we should snapshot the parsed object instead, that way the snapshot will be one several lines and easier to read

src/5.0_to_5.1/migrateUserActivity.test.ts
```diff
@@ -0,0 +1,23 @@
...
+} from "@activeviam/data-model-5.1/dist/__test_resources__";
+
+const dataModels = {
+  sandbox: sandboxDataModel,
+  dualCatalog: dualCatalogDataModel,
+};
+
+describe("migrateUserActivity", () => {
+  it("migrates 5.0 `userFilters` and `userQueryContext` to usable data in 5.1", () => {
+    migrateUserActivity(contentServer, dataModels);
```
eliasinho, line 15: I would add a snapshot of the content before calling `migrateUserActivity` too:
- it's a good illustration of what the function does. When reading the test, one clearly sees the state of the content before the mutation, and after
- it ensures the data from the centralized `contentServer` is not corrupted. It's probably not the scariest example, but let's say someone modifies `user2`'s query context's value from `30` (number) to `"30"` (string), the test would still pass whereas we don't check anymore that the value was stringified

eliasinho (commented):
Looks pretty good! The logic is correct and thanks for adding a test 🙏
I made several comments, but they're only around syntax and phrasing

src/5.0_to_5.1/migrateUserActivity.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const migrateUserActivity = (
+  contentServer: ContentRecord,
+  dataModels: { [serverKey: string]: DataModel },
+): void => {
+  const users = contentServer.children?.ui.children?.users;
+
+  if (users?.children === undefined) {
```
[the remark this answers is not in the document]
dustinstender: Yeah it's cleaner like that

src/5.0_to_5.1/migrateUserActivity.ts
dustinstender: I changed this on my side because I changed the variable names

src/5.0_to_5.1/migrateUserActivity.ts
dustinstender: Did this, but same comment as before, slightly different variable names.

src/5.0_to_5.1/migrateUserActivity.test.ts
dustinstender: you got it boss

src/5.0_to_5.1/migrateUserActivity.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  sandboxDataModel,
+} from "@activeviam/data-model-5.1/dist/__test_resources__";
+
+const dataModels = {
+  sandbox: sandboxDataModel,
+  dualCatalog: dualCatalogDataModel,
+};
+
+describe("migrateUserActivity", () => {
+  it("migrates 5.0 `userFilters` and `userQueryContext` to usable data in 5.1", () => {
```
eliasinho, line 14:
```suggestion
  it("migrates 5.0 user filters and query context to usable ones in 5.1", () => {
```

eliasinho (approved):
Last small thing I forgot in the previous round of review, but otherwise everything looks good to me!

antoinetissier (approved):
Looks good ✅

PR #88, opened 2023-02-07
UI-6616 - Safely migrate user activity

antoinetissier:
Without these safeguards, an error would be thrown if the user activity has no user filters.


antoinetissier:
FYI @dustinstender

PR #98, opened 2023-05-11
UI-8299 - Support notebooks in the Content Migration Tool

mailys-ds:
### Description

The PR is part of UI-8299. It does the following: 

- Add a migration script for Atoti 7.0 notebook, without handling for now the calculated measure (that's why I added an argument to `migrateWidget`.
- Refactor the `src/cli/bin.ts` to call the migration scripts depending on the file extension.
- Add some checks for the different Atoti UI / Atoti version in `bin.ts`.

What need to be done in follow-up PRs:

- [x] Handle case with calculated measures: as decided, we wouldn't handle this case as it is too much of an edge case.
- [x] Add a "migration notebooks" part in the readme: https://github.com/activeviam/atoti-ui-migration/pull/99

### Testing

- Create an Atoti notebook with Atoti 0.7, the AUI version used would be 5.0. Make sure it has a Atoti UI widget, with filters. I can share my test notebook if needed.
- Migrate the notebook with the command line:
 `npx -- activeui-migration -i test.ipynb -o migrated-notebook.ipynb -s servers.json --from-version 0.7 --to-version 0.8`
- Open the migrated notebook in JSON format and check that every widgetState in cells that represent widget have been updated.


eliasinho (commented):
There are a few things that I think won't eventually be in `migrateContentServer` and will stay at the higher level in `bin.ts`.
For example all the part that handles the error reporting can be shared between both scripts I think. That means `counters` and `errorReport` can be defined in `bin.ts` and be passed to the scripts. And all the part that's after `console.log("--------- END OF CONTENT MIGRATION ---------")` can also stay there I think. That also means the `summaryMessages` would stay there.

All that said, if the strategy is to put everything in `migrateContentServer` for now, and once the other script is implemented and inserted in the `bin.ts` logic, factorize what can be factorized between the two scripts, I'm happy with stuff living here for now.
Is that the strategy?

The other way would be to keep the `bin.ts` as it is now. Add all the logic you need directly in it, and do the splitting at the very end. That would avoid moving things now from `bin.ts` that would need to be added back there later. But I don't feel super strongly about either way, so I guess it's more about which strategy you feel more comfortable with.

What do you think?

mailys-ds:
My plan is to move all the script related to the migration of a content in its own file, and then write my migrate notebook scripts in another file and finally merge what is common back in the `bin.ts` as you mentionned in your first bullet point.

src/cli/bin.ts
```diff
@@ -1,91 +1,26 @@
...
-import { getMigrateSavedFilters } from "../getMigrateSavedFilters";
-import { migrate_43_to_50 } from "../4.3_to_5.0";
-import { migrate_50_to_51 } from "../5.0_to_5.1";
-import { getContent } from "../getContent";
+import { migrateContentServer } from "./scripts/migrateContentServer";
+import { migrateNotebook } from "./scripts/migrateNotebook";
+import { getFileType } from "./scripts/getFileType";
 
-const migrationSteps: {
+const AUIMigrationSteps: {
```
messaooudi, line 8: The `AUIMigrationSteps` name is misleading, if I understand correctly, this maps the `from which version` to `to which version` the CLI can operate on. It would be simpler to inline it where its used.

src/cli/bin.ts
```diff
@@ -172,195 +102,48 @@ yargs
...
-      // - in particular, it is the only migration step with async logic
-      if (fromVersion === "4.3") {
-        await migrate_43_to_50(contentServer, {
-          errorReport,
-          counters,
-          servers,
-          keysOfWidgetPluginsToRemove,
-          doesReportIncludeStacks,
+      const fileExtension = getFileExtension(inputPath);
+      const { fromVersion: validFromVersion, toVersion: validToVersion } =
```
messaooudi, line 119: The `fromVersion` and `toVersion` command options has choices attached to them, making them already valid, thus no need to revalidate them.
Also see the `check` call at the end of the file.

messaooudi:
@mailys-ds  What you think about https://github.com/activeviam/atoti-ui-migration/pull/98#discussion_r1205274660

mailys-ds:
@messaooudi Sorry I first did not see your comment.
Also the CI is blocked due to Artifactory connection issues for the moment, that's why it's failing.

src/cli/bin.ts
mailys-ds: I changed the `validateVersions` to convert the potential given Atoti versions in the command line in order to only manipulate AUI versions in the codebase and enable our user to mix Atoti and AtotiUI versions.
Let me know if that works for you.

src/5.0_to_5.1/migrateDashboard.ts
```diff
@@ -40,6 +40,7 @@ export const migrateDashboard: MigrateDashboardCallback<
     dashboardState,
     (widgetState) =>
       migrateWidget(widgetState, {
+        supportCalculatedMeasuresMigration: true,
```
Nouzbe, line 43: I would like to avoid this, and use whether `namesOfCalculatedMeasuresToMigrate` and `measureToCubeMapping` are defined in order to know whether calculated measures should be migrated:

```
type OptionsForCalculatedMeasuresRemoval = {
  namesOfCalculatedMeasuresToMigrate: string[];
  measureToCubeMapping: { [measureName: string]: string[] };
};

/**
  * If calculated measures don't need to be migrated,
  * then the caller does not need to pass options for calculated measures removal.
  */
type Options = OptionsForCalculatedMeasuresRemoval | {};

/**
  * Returns whether `options` allows to remove saved calculated measures from the migrated `widgetState`.
  * This is done when migrating a Content Server, but not a notebook.
  */
function shouldMigrateCalculatedMeasures(options:  Options): options is  OptionsToMigrateCalculatedMeasures {
  return "namesOfCalculatedMeasuresToMigrate" in options && "measureToCubeMapping" in options;
}

export const migrateWidget: MigrateWidgetCallback<
  AWidgetState50,
  AWidgetState51,
  {
    namesOfCalculatedMeasuresToMigrate: string[];
    measureToCubeMapping: { [measureName: string]: string[] };
  } | {}
> = (options) => {
  if (shouldMigrateCalculatedMeasures(options)) {
    migrateCM();
  }
}
```

I prefer doing it this way, because even though it makes the function internals more verbose, it allows a leaner syntax for its callers: they don't have to pass the additional `supportCalculatedMeasuresMigration` argument, and `migrateNotebook` does not have to pass an empty `namesOfCalculatedMeasuresToMigrate` and an empty `measureToCubeMapping` which could otherwise seem obscure to readers who are not yet familiar with `migrateContentServer`.

src/cli/bin.ts
```diff
@@ -1,91 +1,26 @@
...
-    success: "were successfully migrated.",
-    failed:
-      "could not be migrated because errors occurred during their migration.",
-  },
-};
+export function getFileExtension(path: string): string {
+  const fileExtension = path.split(".").pop()?.toUpperCase();
+
+  if (!fileExtension || !supportedFileExtension.includes(fileExtension)) {
+    throw new Error(`File of type ${fileExtension} are not supported.`);
```
Nouzbe, line 19: Let's make the syntax more human and the message more actionable:

```suggestion
    throw new Error(`The extension "${fileExtension}" is not supported. Supported extensions for the input file are "JSON" and "IPYNB"`);
```

src/cli/scripts/convertAtotiToAUIVersions.ts
```diff
@@ -0,0 +1,47 @@
...
+export function convertToVersion(
+  version: typeof possibleToVersions[number],
+): ValidToVersion {
+  if (atotiValidToVersions.includes(version)) {
+    return versionMapping[version] as ValidToVersion;
+  }
+  return version as ValidToVersion;
+}
+
+export function convertVersion(versions: {
```
Nouzbe, line 40: As this converts 2 versions at once.

```suggestion
export function convertVersions(versions: {
```

src/cli/scripts/migrateNotebook.ts
```diff
@@ -0,0 +1,85 @@
...
+        // @ts-expect-error The deserializedWidgetState has been migrated to AWidgetState5.1 where its filters are of type Filter and not MdxExpression.
+        deserializedWidgetState,
+      );
+    }
+  }
+
+  await fs.writeJSON(outputPath, notebook, { spaces: 2 });
+
+  if (numberOfFailures > 0) {
+    console.log(`- ${numberOfFailures} widget(s) have failed to migrate -`);
```
Nouzbe, line 81:
```suggestion
    console.log(`- Could not migrate ${numberOfFailures} widget(s)`);
```

src/cli/scripts/migrateNotebook.ts
```diff
@@ -0,0 +1,87 @@
...
+import { produce } from "immer";
+
+const migrationSteps: {
+  from: string;
+  to: string;
+  migrateWidget: MigrateWidgetCallback<any, any, any>;
+}[] = [{ from: "5.0", to: "5.1", migrateWidget: migrateWidget }];
+
+/**
+ * Migrates the AUI widgets from 5.0 to 5.1 of an Atoti Jupter Notebook.
```
Nouzbe, line 21:
```suggestion
 * Migrates the Atoti UI widgets from 5.0 to 5.1 of an Atoti Jupter Notebook.
```

Nouzbe (commented):
I am almost, amost ready to merge. I would only like to change one thing: the logs. I just tested the notebook and Content Server migration. Both worked, but the logs seem odd to me:

### Content Migration

```
C:\Users\b.amelot\work\activeui-migration>node ./bin -i content.json -o migrated.json -s servers.json --from-version 4.3 --to-version 5.1
--------- START OF CONTENT MIGRATION ---------
--------- END OF CONTENT MIGRATION ---------

# Dashboards
- 1060 were partially migrated, but errors occurred in some of the widgets they contain. These widgets were copied as is into the migrated dashboards.
- 581 could not be migrated because errors occurred during their migration. Their original versions were copied as is into the migrated folder.
- 18 were cleaned up because they could not be found in the ui/dashboards/structure folder. They were already not visible in your version of ActiveUI.

# Widgets
- 4 were migrated with warnings.
- 175 could not be migrated because errors occurred during their migration. Their original versions were copied as is into the migrated folder.
- 3 were cleaned up because they could not be found in the ui/widgets/structure folder or because their keys were passed in the --remove-widgets option.

# Filters
- 11 were successfully migrated.
- 11 could not be migrated because errors occurred during their migration. Their original versions were copied as is into the migrated folder.

# Folders
- 1 were cleaned up because they could not be found in their structure folder. They were already not visible in your version of ActiveUI.


For more information about the errors that occurred, rerun the command with the `--debug` option.
This will output a file named `report.json` containing the error messages.
To see the stack traces of the errors in this file, you can also use the `--stack` option.
--------------------------------------------
```

I would remove the first line: the one with "Start of content migration". I believe that's how it was before this PR. I think it's odd to me, because there is no line of log between ----start---- and ----end----.


### Notebook migration

```
C:\Users\b.amelot\work\activeui-migration>node ./bin -i test.ipynb -o migrated.ipynb -s servers.json --from-version 0.7 --to-version 0.8
--------- START OF NOTEBOOK MIGRATION ---------
--------- END OF NOTEBOOK MIGRATION ---------

Here, I would remove these two lines and just replace it with one containing information. I am thinking of:

```
- Successfully migrated n wigets.
(2nd line only if applicable) - Failed to migrate m widgets.
```

What do you think?

Nouzbe (approved):
Let's go! 🏁

PR #100, opened 2023-08-04
add tarballs and update package.json and readme for SCB to build offline

mingkongwoo:
@tht-activeviam can you check if anything missing before merge to scb-5.0.9.
Thanks


tht-activeviam (approved):
Tested in my machine and it is working.

PR #101, opened 2023-08-11
UI-8677 - Handle disconnected widgets

messaooudi:
In AU4, a widget can be disconnected from its page and dashboard filters.
In Atoti UI 5, that feature does not exist. The migration of it goes as follow:
 - If the dashboard contains a disconnected widget, then move the dashboard filters down to every page.
 - If a given page contains a disconnected widget, then move the page filters down to every connected widget.

### Note
I had to regenerate the `yarn.lock` file because i think it was corrupted, the reason for that is:
- in the `package.json` file, we specify that the `@activeviam` dependencies should be pulled from the local vendor folder, whereas in `yarn.lock` they were still tight to the the private artifactory url.


package.json
```diff
@@ -22,7 +22,7 @@
   ],
   "scripts": {
     "start": "node bin.js",
-    "build": "webpack",
+    "build": "cross-env NODE_OPTIONS=--openssl-legacy-provider webpack",
```
messaooudi, line 25: this is suggested in  `README_SCB_MarketRisk.md`

src/migrateDashboard.ts
```diff
@@ -92,9 +94,21 @@ export function migrateDashboard(
   body.pages.forEach((legacyPage: LegacyDashboardPage, index: number) => {
     const pageKey = `p-${index}`;
     const content: DashboardPageState<"serialized">["content"] = {};
+    keysOfDisconnectedWidgets[pageKey] = [];
     legacyPage.content.forEach((widget) => {
       const leafKey = widget.key;
       const widgetPluginKey = _getLegacyWidgetPluginKey(widget.bookmark);
+
+      if (
+        widget.bookmark.value.body?.configuration?.tabular?.isConnected ===
```
Nouzbe, line 103: I suspect that `isConnected` might have been set in this place for table widgets but probably not for other types of widgets. We should check whether this information always lives there or not.

src/migrateDashboard.ts
```diff
@@ -180,6 +194,30 @@ export function migrateDashboard(
...
+    delete dashboard.filters;
+
+    Object.entries(pages).forEach(([pageKey, page]) => {
+      page.filters = [...(dashboardFilters ?? []), ...(page.filters ?? [])];
+      if (!_isEmpty(keysOfDisconnectedWidgets[pageKey])) {
+        const pageFilters = page.filters;
+        delete page.filters;
+
+        Object.entries(page.content).forEach(([leafKey, widget]) => {
+          if (!keysOfDisconnectedWidgets[pageKey].includes(leafKey)) {
```
Nouzbe, line 210: Nitpicky, but this could be made more performant if `keysOfDisconnectedWidgets` contained sets instead of arrays.

PR #102, opened 2023-08-11
UI-8712 - Bump `@activeviam` dependencies to `5.0.27-20230811125028-dfec9c3` & Proper offline install

messaooudi:
Does what the title says


Nouzbe:
The people who were using this branch did not report to me that it was broken, so I am not sure if this will fix it or break it for them. Therefore I prefer to close it for now. As discussed together, we will discuss in September which commits to move to `main` after this SCB migration. It's the only thing that matters. We don't have to care about this branch.

PR #104, opened 2023-08-14
UI-8671 - (SCB Migration) Adjust column width to accomodate for deep level expansion

NZepeda:
## Summary
Part of the SCB migration. This PR introduces a new cli argument `tree-column-width` which accepts two numbers separated by a comma to use for calculating the width of the first column. 

Example: `--tree-column-width 200,50` will result in a calculation of 200 + (50 * maxLevelDepth). See the result in the UI in the images below.

## Testing
**Note 1**: Ideally you can spin up a local activepivot server in order to test otherwise you can use https://activepivot-ranch.activeviam.com:5110

**Note 2**: Before doing `yarn && yarn build` make sure your environment is set to use node `v16.x.x`. I'm using `v16.14.2` and that works well.

0. Start your local activepivot server, you'll need this later for uploading and testing the migrated content.
1. Check this branch out locally
2. Within the repo run `yarn && yarn build`
3. Within the repo, create a folder named `data` at the root of the project.
4. Create a new file within the `data` folder and name it `content.json`
5. Paste this [json](https://api.jsonbin.io/v3/b/64de29b39d312622a392a5d0) within that file.
6. Create another new file within the `data` folder and name `servers.json`.
7. Paste this [json](https://api.jsonbin.io/v3/b/64de2ab88e4aa6225ed1727c) within that file.
8. Within the repo directory, run the script with the following command: 
```
npm start -- --debug --input-path ./data/content.json --servers-path ./data/servers.json --output-path migrated.json --tree-column-width 200,50 
```
9. Open up your local activepivot server and navigate to the admin UI. It should be at something like `http://localhost:9090/admin/ui/index.html`.
10. In the `content` panel of admin UI, make sure the `UI` exists. 
11. Right-click the UI folder and click `import`. 
12. Select the newly created `migrated.json`
13. Finally, using a UI (make sure you're using AUI-5) pointing to this content server, open the dashboards. (**compositeHierarchy** and **regular hierarchy**).
**Expected:** Both dashboards should have their first column resized

### Before
dashboard-no-column-width

### After
Screenshot 2023-08-14 at 11 39 19 AM

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -30,9 +30,10 @@
     lodash "^4.17.15"
     tslib "^2.0.0"
 
-"@activeviam/activeui-sdk-scripts@file:vendor/activeviam-activeui-sdk-scripts-5.0.27-20230803200406-61fc83c.tgz":
-  version "5.0.27-20230803200406-61fc83c"
-  resolved "file:vendor/activeviam-activeui-sdk-scripts-5.0.27-20230803200406-61fc83c.tgz#9cd2590e18216f6f130aa7b20ac7a678121dabe0"
+"@activeviam/activeui-sdk-scripts@5.0.27-20230811125028-dfec9c3":
+  version "5.0.27-20230811125028-dfec9c3"
+  resolved "https://activeviam.jfrog.io/artifactory/api/npm/npm-internal/@activeviam/activeui-sdk-scripts/-/@activeviam/activeui-sdk-scripts-5.0.27-20230811125028-dfec9c3.tgz"
+  integrity sha512-SKgmuCF/FvOoxGOh4385UFZ7PM4Ku8QRkez8+StxGdjrJFnCiwEbtkdrfkyqlB+woTLuqKF0kln+bF+pbsd17w==
   dependencies:
     "@activeviam/content-server-initialization" "5.0.27-20230811125028-dfec9c3"
     "@babel/cli" "^7.11.6"
@@ -93,9 +94,10 @@
     webpack-manifest-plugin "^3.1.0"
     yargs "^15.3.1"
 
-"@activeviam/activeui-sdk@file:vendor/activeviam-activeui-sdk-5.0.27-20230803200406-61fc83c.tgz":
-  version "5.0.27-20230803200406-61fc83c"
-  resolved "file:vendor/activeviam-activeui-sdk-5.0.27-20230803200406-61fc83c.tgz#9341a37752b0bcca1687ab20755e3ec4d0eaafcf"
+"@activeviam/activeui-sdk@5.0.27-20230811125028-dfec9c3":
+  version "5.0.27-20230811125028-dfec9c3"
+  resolved "https://activeviam.jfrog.io/artifactory/api/npm/npm-internal/@activeviam/activeui-sdk/-/@activeviam/activeui-sdk-5.0.27-20230811125028-dfec9c3.tgz"
+  integrity sha512-nATxOpMZ3wGvKw29OfSPwZCqsHSoHm1B8DoISU54meDy3Pl2mMCOQI//UZ06lg6Sz2zh9lUYjeAm8u/NMIVP8Q==
   dependencies:
     "@activeviam/action" "5.0.27-20230811125028-dfec9c3"
     "@activeviam/activepivot-client" "5.0.27-20230811125028-dfec9c3"
@@ -390,9 +392,10 @@
     lodash "^4.17.15"
     tslib "^2.0.0"
 
-"@activeviam/content-server-initialization@5.0.27-20230803200406-61fc83c", "@activeviam/content-server-initialization@file:vendor/activeviam-content-server-initialization-5.0.27-20230803200406-61fc83c.tgz":
-  version "5.0.27-20230803200406-61fc83c"
-  resolved "file:vendor/activeviam-content-server-initialization-5.0.27-20230803200406-61fc83c.tgz#80aa703ba6ae46a5aeba12b5826fa7e90a0326f7"
+"@activeviam/content-server-initialization@5.0.27-20230811125028-dfec9c3":
+  version "5.0.27-20230811125028-dfec9c3"
+  resolved "https://activeviam.jfrog.io/artifactory/api/npm/npm-internal/@activeviam/content-server-initialization/-/@activeviam/content-server-initialization-5.0.27-20230811125028-dfec9c3.tgz"
+  integrity sha512-cZOEly4bDvRSlT8iERa+KzaonBz6NyC0Xn3OUOzsg0c1/3kDCwPKW2VGDgxC25rFDenC8Yh9dOH2hoC4wh7a3w==
   dependencies:
     isomorphic-unfetch "^3.1.0"
     lodash "^4.17.15"
@@ -489,18 +492,10 @@
     lodash "^4.17.15"
     ts-essentials "^6.0.5"
 
-"@activeviam/data-model@file:vendor/activeviam-data-model-5.0.27-20230803200406-61fc83c.tgz":
-  version "5.0.27-20230803200406-61fc83c"
-  resolved "file:vendor/activeviam-data-model-5.0.27-20230803200406-61fc83c.tgz#9a8c374a359e4416229f1bc6f42b5c45d7ae1a00"
-  dependencies:
-    "@types/lodash" "^4.14.168"
-    lodash "^4.17.15"
-    ts-essentials "^6.0.5"
-
-"@activeviam/data-visualization@5.0.27-20230803200406-61fc83c":
-  version "5.0.27-20230803200406-61fc83c"
-  resolved "https://activeviam.jfrog.io/artifactory/api/npm/npm-internal/@activeviam/data-visualization/-/@activeviam/data-visualization-5.0.27-20230803200406-61fc83c.tgz#2a7d0e2ba963f881f71d0ceb5efb3d1215912c3c"
-  integrity sha512-G4IYQGsYUFXg2U2zc66mT/Jy66y2OQNeasnV6LF/Aa5iFF5/cohgI29Hd4D1xRHC/EST6MsPUvq8BRfWhAjAiw==
+"@activeviam/data-visualization@5.0.27-20230811125028-dfec9c3":
+  version "5.0.27-20230811125028-dfec9c3"
+  resolved "https://activeviam.jfrog.io/artifactory/api/npm/npm-internal/@activeviam/data-visualization/-/@activeviam/data-visualization-5.0.27-20230811125028-dfec9c3.tgz"
+  integrity sha512-EUF6No+JZEkL5NQrCJuxYDuMQVzZfJDKFn/9ZjINZkkKFvrq4obGgzEe/o/imRsdGgEPfOUxaimM8M+Zf4E/6Q==
   dependencies:
     "@activeviam/action" "5.0.27-20230811125028-dfec9c3"
     "@activeviam/calculated-measures" "5.0.27-20230811125028-dfec9c3"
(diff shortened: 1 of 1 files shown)


src/migrateTable.ts
```diff
@@ -65,11 +67,17 @@ export function migrateTable(
...
+
+  const columnsAxis = query.mdx?.axes[0];
+  const columnLevels = columnsAxis ? getLevels(columnsAxis, { cube }) : [];
+
   const columnWidths = legacyColumns
     ? _migrateTableColumnWidths({
         legacyColumns,
         mapping,
         cube,
+        maxLevelDepth: columnLevels.length,
```
Nouzbe, line 79: I believe that the computation of the max level depth is a bit more complex as it should include the index of each level within its parent hierarchy. For instance in the sandbox, if `columnLevels` contains a single element which is `BookId`, then the max level depth is 4 and not 1. Indeed, the column contains legal entities, business units, desks and finally book ids.

src/cli/bin.ts
```diff
@@ -45,6 +46,7 @@ yargs
     pivotInputPath?: string;
     debug: boolean;
     stack: boolean;
+    columnWidth?: string;
```
Nouzbe, line 49: I would prefer a more specific name. Otherwise it's ambiguous which column we are refering to.

```suggestion
    treeColumnWidth?: string;
```

src/cli/bin.ts
```diff
@@ -78,6 +80,13 @@ yargs
         demandOption: false,
         desc: "The path to the JSON export of the /pivot folder on the content server.",
       });
+      args.option("column-width", {
```
Nouzbe, line 83: Same suggestion here.

dustinstender (commented):
Just so I understand correctly, this is just to make it so that SCB doesn't have to manually expand columns for each of their tree table widgets?

Also, the `baseWidth` is used to represent what a column width should be for a single member and then the `levelMultiplier` is used to add width when showing the levels leading to a specific level (like in the photo you posted). Is this correct?

src/getMaxLevelDepthFromMapping.ts
```diff
@@ -0,0 +1,79 @@
...
+            )
+          : undefined;
+
+      const { dimensionName, hierarchyName } = mapping.rows[0];
+
+      const levelName = firstLevel
+        ? firstLevel.name
+        : mapping.rows[0].levelName;
+
+      columnKey = quote(dimensionName, hierarchyName, levelName);
```
NZepeda, line 59: This may be due to a potential bug on Atoti UI 5, but I found that if the first column is represented by a `regular` hierarchy (non-composite), and that hierarchy expanded down to a deeper level than the first, for example, the column is representing  `[Booking].[Desk].[BookId]`, its column key in the `columnWidths` property of the widget metadata  is represented by the first level of the hierarchy, in this case `[Booking].[Desk].[LegalEntity]`. To me this seems like a bug on the UI end since it would make more sense that the column should be represented by the more accurate `[Booking].[Desk].[BookId]`. This is to accommodate for that discrepancy.

src/migrateTable.ts
NZepeda: My latest commits address this comment 👍

src/cli/bin.ts
NZepeda: Done!

src/cli/bin.ts
NZepeda: Done!

src/getMaxLevelDepthFromMapping.ts
```diff
@@ -0,0 +1,79 @@
...
+  treeTableColumnWidth,
+  columnLevels,
+  cube,
+}: {
+  mapping: DataVisualizationWidgetMapping;
+  maxLevelDepth: number;
+  treeTableColumnWidth: [number, number];
+  columnLevels: LevelCoordinates[];
+  cube: Cube;
+}): { [columnKey: string]: number } {
```
dustinstender, line 25: Why do we call this `columnKey` instead of `columnWidth`?

dustinstender (commented):
This looks pretty good to me. I'm just curious about how we are moving forward. Will SCB just test this out and if it's good for them we just go ahead and merge this basically?

src/getMaxLevelDepthFromMapping.ts
NZepeda: Looking at this more closely, I'm going to push another commit to rename this file. I renamed the function but not the file name 🤦‍♂️

dustinstender: Okay thanks for the explanation

NZepeda:
@dustinstender I've added one more commit to this branch, can you have another look?

dustinstender:
Hey, while going through the testing procedure I ran into this issue while running the command you posted.

Screenshot 2023-08-17 174505

Any idea why? I ran `yarn && yarn build`.

src/getTreeColumnWidthFromArgs.ts
```diff
@@ -0,0 +1,21 @@
...
+    .split(",")
+    .map((stringValue) => parseInt(stringValue));
+
+  if (values.length !== 2) {
+    throw new Error(
+      "The value passed into the column-width argument must be two numbers separated by a comma",
+    );
+  }
+
+  const [baseWidth, levelColumnMultiplier] = values;
```
Nouzbe, line 18: Let's align the name with the one you give it in the multiline comment above the function.

```suggestion
  const [baseWidth, levelDepthMultiplier] = values;
```

src/migrateTable.ts
```diff
@@ -65,13 +69,24 @@ export function migrateTable(
...
     legacyTableState.value?.body?.configuration?.tabular?.columns;
-  const columnWidths = legacyColumns
-    ? _migrateTableColumnWidths({
-        legacyColumns,
-        mapping,
-        cube,
-      })
-    : {};
+
+  const columnsAxis = query.mdx?.axes[0];
```
Nouzbe, line 73: I believe that you are referring to the maximum level depth on rows here.

```suggestion
  const rowsAxis = query.mdx?.axes[0];
```

Also, this will only work if their query has the `ON ROWS` clause before the `ON COLUMNS` one. If some of their queries don't follow this rule, then the migration will yield unexpected results regarding this width.

Instead, we should do something like:

```
query.mdx.axes.find(axis => axis.name === "ROWS");
```
But note that `axis.name` can actually be different things: the case can vary and it can be a number too. See [UI-8726](https://activeviam.atlassian.net/browse/UI-8726) for a recent example of related bug.

src/migrateTable.ts
```diff
@@ -65,13 +69,24 @@ export function migrateTable(
...
+  const columnLevels = columnsAxis ? getLevels(columnsAxis, { cube }) : [];
+
+  const maxLevelDepth = _max(
```
Nouzbe, line 76: We need a sum here, not a max. Indeed, if the table has Desk (3rd level in its hierarchy) and Counterparty (2nd level in its hierarchy) on rows, then a row in the tree column will typically contain 5 elements and therefore the depth is 5:

| LegalEntityA | BusinessUnitA | DeskA | Asahi | Asahi Chemical Ind. |
|---------------|-----------------|--------|--------|-----------------------|

src/migrateTable.ts
Nouzbe: There is another possible way and it is to derive the level depth from the mapping instead of the query, it might be simpler.

src/getTreeColumnWidth.ts
```diff
@@ -0,0 +1,79 @@
...
+  LevelCoordinates,
+  getHierarchy,
+  getLevel,
+} from "@activeviam/activeui-sdk";
+import { quote } from "@activeviam/mdx";
+
+/**
+ * Returns the column width for the tree table calculated by the baseWidth + (maxLevelDepth * levelMultiplier)
+ */
+export function getTreeColumnWidth({
```
Nouzbe, line 13: It would be nice to rename this function into `getTreeColumnKey`, make it only responsible for returning the key and not the width and inline the computation of the width where it is needed, since it's a one liner and all the rest of this function here is concerned about the key of the tree column, not its width.

src/getTreeColumnWidth.ts
```diff
@@ -0,0 +1,79 @@
...
+        ? getHierarchy(
+            {
+              dimensionName: columnLevels[0].dimensionName,
+              hierarchyName: columnLevels[0].hierarchyName,
+            },
+            cube,
+          )
+        : undefined;
+
+      const firstLevelName = hierarchy && hierarchy.levels[1];
```
Nouzbe, line 40: This will probably cause a bug in case `hierarchy` is slicing.

src/getTreeColumnWidth.ts
```diff
@@ -0,0 +1,79 @@
...
+  maxLevelDepth,
+  mapping,
+  treeTableColumnWidth,
+  columnLevels,
+  cube,
+}: {
+  mapping: DataVisualizationWidgetMapping;
+  maxLevelDepth: number;
+  treeTableColumnWidth: [number, number];
+  columnLevels: LevelCoordinates[];
```
Nouzbe, line 23: This should be called `rowLevels` (same remark as above). I also believe that this information is redundant with `mapping.rows` and that we should remove one of these two arguments.

If `rowLevels` contains the measures, then I believe that using it is simpler. But if it does not contain this information, then we should use `mapping.rows`.

I might be wrong, but if we base ourselves on `rowLevels`, then I believe that the function boils down to:

```
if (rowLevels.length === 0) {
  return undefined;
}

const { dimensionName, hierarchyName } = rowLevels[0];

if (dimensionName === "Measures") {
  return "[Measures].[Measures]";
}

const hierarchy = getHierarchy({ dimensionName, hierarchyName });
const firstLevel = hierarchy.levels[hierarchy.slicing ? 0 : 1];
return quote(dimensionName, hierarchyName, firstLevel.name);
```

src/getTreeColumnWidth.ts
```diff
@@ -0,0 +1,79 @@
...
+        mapping.rows[0].hierarchies[0];
+      columnKey = quote(dimensionName, hierarchyName, levelName);
+      break;
+    }
+    case "allMeasures": {
+      columnKey = "[Measures].[Measures]";
+      break;
+    }
+    default: {
+      throw new Error("A measure cannot be mapped on an ordinal field.");
```
Nouzbe, line 73: This looks like an unfortunate copy🍝

src/getTreeColumnWidth.ts
NZepeda: Done 👍

dustinstender (approved):
I was able to follow the steps and indeed the first column is resized. Lookin' good

Nouzbe (commented):
@NZepeda I pushed a commit attempting to simplify the process of retrieving the total level depth on rows, because I was struggling to read it at first.

I wrote a few tests but I admit I did not run them yet.

Can you let me know what you think / whether tests pass?

src/_getLevelDepthOnRows.test.ts
```diff
@@ -0,0 +1,82 @@
...
+          type: "hierarchy",
+          dimensionName: "Time",
+          hierarchyName: "HistoricalDates",
+          levelName: "AsOfDate",
+        },
+      ],
+    };
+
+    // Desk is the 3rd level on the [Booking].[Desk] hierarchy (which is non slicing).
+    // AsOfDate is the 1rst one on the [Time].[HistoricalDates] hierarchy (which is slicing).
```
NZepeda, line 27:
```suggestion
// AsOfDate is the 1st level on the [Time].[HistoricalDates] hierarchy (which is slicing).
```

NZepeda:
> I pushed a commit attempting to simplify the process of retrieving the total level depth on rows, because I was struggling to read it at first.
> 
> I wrote a few tests but I admit I did not run them yet.
> 
> Can you let me know what you think / whether tests pass?

@Nouzbe I just looked over your commit and it looks good. Good simplification, your new tests pass while the tests I had previously added continue to pass as well 👍

src/getTreeColumnKey.ts
```diff
@@ -0,0 +1,65 @@
...
+            )
+          : undefined;
+
+      const { dimensionName, hierarchyName } = mapping.rows[0];
+
+      const levelName = firstLevel
+        ? firstLevel.name
+        : mapping.rows[0].levelName;
+
+      return quote(dimensionName, hierarchyName, levelName);
```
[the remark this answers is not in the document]
NZepeda: @Nouzbe Indeed it can and this is actually what I was referring to [here](https://github.com/activeviam/atoti-ui-migration/pull/104#discussion_r1294170228). If you have the first column representing a level that is **not** the first level of the hierarchy (see image below), `bookId` for example, instead of the column key being `[Booking].[Desk].[BookId]`, AUI 5 expects it to be `[Booking].[Desk].[LegalEntity] ([Dimension].[Hierarchy].[firstLevelOfHierarchy])`, which is why I've added a bit more logic to make sure the column key ends up being what AUI 5 expects. 

Screenshot 2023-08-22 at 4 09 59 PM

Nouzbe (approved):
Looks good 👍 

You were planning on removing `data/content.json` and `data/servers.json` before merging, right? If that's so, don't forget to do it and feel free to merge!

PR #109, opened 2023-08-28
UI-8671 - Integrate SCB tree table column migration into the core

NZepeda:
## Summary
This PR cherry-picks the changes done for the SCB migration that allows them to adjust the size of the column on tree tables

[**Original PR**](https://github.com/activeviam/atoti-ui-migration/pull/104)

### DIFF
--- a/src/_getLevelDepthOnRows.ts
+++ b/src/_getLevelDepthOnRows.ts
@@ -1,0 +1,80 @@
+import _sum from "lodash/sum";
+import {
+  DataVisualizationWidgetMapping,
+  HierarchyInMapping,
+  quote,
+} from "@activeviam/activeui-sdk";
+import {
+  Cube,
+  LevelCoordinates,
+  getHierarchy,
+  getLevelIndex,
+} from "@activeviam/data-model";
+
+/**
+ * Returns the depth of the level corresponding to the given `levelCoordinates`.
+ * {@example}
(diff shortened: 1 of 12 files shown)


src/_getLevelDepthOnRows.ts
antoinetissier, line 1: This file and the corresponding test file are used only in the migration from 4.3 to 5.0 right ?
If so le's move them there.

antoinetissier (commented):
The cherry-pick looks good, however I am curious whether this solution was discussed directly with the client ?
It seems odd to me to specify the same base width for all tables.
And independently from that, on how many dashboards to they have the problem ? Is it just a few or hundreds ?

NZepeda:
> The cherry-pick looks good, however I am curious whether this solution was discussed directly with the client ? It seems odd to me to specify the same base width for all tables. And independently from that, on how many dashboards to they have the problem ? Is it just a few or hundreds ?

@antoinetissier I believe @Nouzbe discussed this more directly with the client, but because of the complexities of correctly calculating the width properly for all tree tables, we went with a more heuristic fix of allowing them to use the `--tree-column-width` argument to pass in whatever base width is necessary. You can see my comment on the ticket about it [here](https://activeviam.atlassian.net/browse/UI-8671?focusedCommentId=1084934).

antoinetissier (approved):
Sounds good, let's go 🏁

PR #114, opened 2023-09-14
UI-8532 - Take hidden columns into account during table migrations from 4.3 to 5.0

Nouzbe:
Before this PR, hidden columns in AUI 4.3 would not be hidden post-migration, when the table is viewed in AUI 5.0.

This PR fixes it.


antoinetissier (approved):
All good 💯

PR #115, opened 2023-09-25
UI-8742 - More robust MDX cleanup

antoinetissier:
- Bump to the latest versions of the SDK in order to benefit from the MDX changes implemented recently
- In particular, use `getIndexOfDeepestLevelExpressedInDescendantsNode` which has been improved on the SDK side, and allows not to reinvent the wheel here
- Add a try/catch around the different MDX cleanup steps, because they should not be blocking if they fail


src/4.3_to_5.0/_cleanupDescendants.test.ts
```diff
@@ -60,7 +60,7 @@ describe("_cleanupDescendants", () => {
       NON EMPTY Hierarchize(
             Descendants(
               {
-                [Booking].[Desk].[ALL].[AllMember].[LegalEntityA]
+                [Booking].[Desk].[ALL].[AllMember].[LegalEntityA].[BusinessUnitA].[DeskA].[0]
```
antoinetissier, line 63: We were actually checking something invalid with the previous input: 
``` 
Descendants(
              {
                [Booking].[Desk].[ALL].[AllMember].[LegalEntityA]
               },
               1
)
```
should NOT be replaced by just `LegalEntityA`.
Indeed the level at distance 1 from `LegalEntityA` is `BusinessUnit`.

However in the new test input:
``` 
Descendants(
              {
                [Booking].[Desk].[ALL].[AllMember].[LegalEntityA].[BusinessUnitA].[DeskA].[0]
               },
               1
)
```
the Descendants node should be replaced by the `0` BookId, since it doesn't have descendants anyway.

src/4.3_to_5.0/_cleanupDrilldownLevel.ts
```diff
@@ -17,15 +21,26 @@ const canDrilldownLevelBeReplacedByItsFirstArgument = (
   drilldownLevelNode: MdxFunction,
   cube: Cube,
 ) => {
```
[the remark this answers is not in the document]
antoinetissier: Replace if and only if both these conditions are met:
- the function has a single argument
- in the input set, the deepest level expressed in the first hierarchy (which is the one implicitly drilled one) is a leaf level

src/4.3_to_5.0/_cleanupDrilldownLevel.ts
```diff
@@ -11,21 +15,34 @@ import { produce } from "immer";
 import _set from "lodash/set";
 
 /**
- * If the input set holds members of the leaf level, then DrilldownLevel can be replaced by it.
+ * A DrilldownLevel can be replaced by its first argument if both the following conditions are met:
+ * - It has a single argument
```
Nouzbe, line 19:
```suggestion
 * - It has a single argument, which is a set.
```
Otherwise, the _"this set"_ is ambiguous on the next line.

src/4.3_to_5.0/_cleanupDrilldownLevel.ts
```diff
@@ -11,21 +15,34 @@ import { produce } from "immer";
...
+ * - In this set, the deepest level expressed in the first hierarchy (which is the one being implicitly drilled) is a leaf level
```
Nouzbe, line 20:
```suggestion
 * - In this set, the deepest level expressed in the first hierarchy (which is the one being implicitly drilled) is a leaf level (and therefore cannot be drilled down)
```

Nouzbe (approved):
Looks good 👍 
You can consider this couple suggestions:

PR #119, opened 2023-10-05
UI-8859 - Fix error thrown when all widgets but one are removed from a page

Nouzbe:
This is a blocker for the Santander upgrade (from 4.3 to 5.1).
The error is not mentioned in the ticket, but the scenario is well enough described that I could reproduce it with a unit test.

The cause is that during the 4.3 -> 5.0 migration, we would not add anything under `page.content` for widgets that would be removed anyway. The problem is that the widget removal logic depends on it being present both in layout and [in content](https://github.com/activeviam/atoti-ui/blob/6d19eadb48bd756280c0208f078ed8805ad873b4/packages/dashboard-base/src/utils/removeWidget.ts#L67).

This PR fixes the bug and should unlock the Santander upgrade.

### DIFF
--- a/src/4.3_to_5.0/migrateDashboard.test.ts
+++ b/src/4.3_to_5.0/migrateDashboard.test.ts
@@ -311,6 +311,43 @@ describe("migrateDashboard", () => {
     `);
   });
 
+  it("does not throw when all widgets but one are removed from the page", () => {
+    const keysOfWidgetPluginsToRemove = ["filters", "quick-filter", "chart"];
+
+    // Safeguard to make sure that the test makes sense: before checking that the key has been removed in the migrated dashboard, check that it's here in the first place in the legacy dashboard.
+    const widgetPluginKeysInLegacyDashboard: string[] = [];
+    legacyDashboard.value.body.pages.forEach((page) =>
+      page.content.forEach(({ bookmark }) =>
+        widgetPluginKeysInLegacyDashboard.push(bookmark.value.containerKey),
+      ),
+    );
+    expect(widgetPluginKeysInLegacyDashboard).toEqual(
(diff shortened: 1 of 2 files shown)


antoinetissier:
Nice for the quick fix 👍 

Personally, I would have done it on the `atoti-ui` side though, in `removeWidget`: [here](https://github.com/activeviam/atoti-ui/blob/6d19eadb48bd756280c0208f078ed8805ad873b4/packages/dashboard-base/src/utils/removeWidget.ts#L67), add an extra condition that the removed leaf key is indeed part of `Object.keys(content)`.

What do you think ?
I'm approving here anyway as it cannot hurt.

antoinetissier (approved):

antoinetissier:
Also does it close this ? https://github.com/activeviam/atoti-ui-migration/pull/44

Nouzbe:
> Personally, I would have done it on the atoti-ui side, in removeWidget: [here](https://github.com/activeviam/atoti-ui/blob/6d19eadb48bd756280c0208f078ed8805ad873b4/packages/dashboard-base/src/utils/removeWidget.ts#L67), add an extra condition that the removed leaf key is indeed part of Object.keys(content).

I don't find the `removeWidget` buggy. It expects the `pageState` to be in a consistent state: widgets present in `pageState.content` should also be in `pageState.content` (and the other way around too).

The way I see it, the migration script was misusing this function by doing part of the job first (removing from `pageState.content`) and then calling `removeWidget` to do the (more difficult) part of the work: removing from `pageState.layout`.

Nouzbe:
> Also does it close this ? https://github.com/activeviam/atoti-ui-migration/pull/44

I believe so, yes.

PR #120, opened 2023-10-12
UI-8803 - Support namespaces in XML calculated members

antoinetissier:
See UI-8803, and the added test: the problem is that the `calculatedMember` tag in the customer's XML is prefixed with a namespace (`ns2`).
Adding a [tag processor](https://www.npmjs.com/package/xml2js#processing-attribute-tag-names-and-values) allows to remove it from the parsed JSON.

Note that when migrating all the way from 4.3 to 5.1, then:
- during 4.3 to 5.0, the prefix is removed
- it it not added again in 5.0 to 5.1

I honestly think it's fine, and would prefer for clients to report an issue before trying to preserve it.
For example, in this example from the customer, [`ns2` is actually a dummy namespace that should only be used in test cases](https://www.w3.org/XML/2008/xsdl-exx/ns2).


Nouzbe (approved):
👍

PR #122, opened 2023-12-13
UI-9025 - Fix command snippets in README.md

Nouzbe:
Half our README snippets don't work, causing SocGen and Santander to not manage to use our migration script without help from ActiveViam CS.

This PR fixes it.


antoinetissier (approved):
Nice, thank you for the fix.

PR #123, opened 2024-03-14
UI-9249 - Migrate default selected drillthrough columns

Nouzbe:
The migration script used to disregard default selected drillthrough columns configured in UI 4.3 when it is used to migrate the content to a more recent version.

This PR takes care of it and makes sure that this configuration is not lost.

### DIFF
--- a/src/4.3_to_5.0/legacySettingsFolder.ts
+++ b/src/4.3_to_5.0/legacySettingsFolder.ts
@@ -14,6 +14,58 @@ const user1Preferences = {
   deny: [],
   map: {
     "userFilters.enabled": true,
+    "widgets.Tabular.drillthrough.selectedColumns": {
+      "https://activerisk.zone1.scb.net:8081/server": {
+        MarketRiskCube: [
+          {
+            functionName: "MemberValue",
+            columnName: "riskCurrency",
+          },
+          {
+            functionName: "Caption",
+            columnName: "riskCurrency",
+          },
+          {
+            functionName: "MemberValue",
+            columnName: "subTradeId",
+          },
+          {
+            functionName: "Caption",
+            columnName: "subTradeId",
+          },
+        ],
+        SecondCube: [
+          {
+            functionName: "MemberValue",
+            columnName: "delta",
+          },
+          {
+            functionName: "Caption",
+            columnName: "delta",
+          },
+        ],
+      },
+      "https://second-server.com": {
+        MarketRiskCube: [
+          {
+            functionName: "MemberValue",
+            columnName: "subTradeId",
+          },
+          {
+            functionName: "Caption",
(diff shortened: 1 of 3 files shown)


src/4.3_to_5.0/migrateSettingsFolder.ts
```diff
@@ -38,6 +39,43 @@ function migrateSettingsMap(legacySettingsMap: {
...
+        );
+      }
+    }
+
+    // Remove duplicates.
+    migratedSettingsMap["drillthrough.defaultSelectedColumns"] = _mapValues(
+      migratedDrillthroughColumns,
+      _uniq,
+    );
+  }
```
messaooudi, line 77: This implies that two drillthrough columns with the same `columnName`  but different `functionName` are considered duplicates. Right ?

src/4.3_to_5.0/legacySettingsFolder.ts
```diff
@@ -14,6 +14,58 @@ const user1Preferences = {
...
+        ],
+        SecondCube: [
+          {
+            functionName: "MemberValue",
+            columnName: "delta",
+          },
+          {
+            functionName: "Caption",
+            columnName: "delta",
+          },
```
messaooudi, line 45: This is considered duplicates, so we are testing that part of the implementation ?

messaooudi (commented):
Looks good, I have one question to make sur i understand the remove duplicate part.

src/4.3_to_5.0/migrateSettingsFolder.ts
Nouzbe: Indeed. In fact there are even two ways to have duplicates:
- because UI4 was typically storing the following for each column:
```
{
  functionName: "MemberValue",
  columnName: "myColumn",
}, {
  functionName: "MemberCaption",
  columnName: "myColumn"
}
```
So we don't want to save each column twice in the new Atoti UI 5 setting.

The other way is more rare. It is because UI4 was saving the columns per server and per cube whereas UI5 only stores them per cube. So you could have a duplicate because a column was save for a cube with the same name under two different servers.

The current implementation in this PR should get rid of any duplicate, whether they come from the first way or the second one I described here.

src/4.3_to_5.0/legacySettingsFolder.ts
Nouzbe: This is considered duplicates indeed, but it's the first way to have duplicate I described above. Technically, I didn't test the second way of having duplicates.

I also tested the second way of having duplicates, as I added the column `subTradeId` in `MarketRiskCube` under two different servers.

messaooudi (approved):
Thanks for the explanations. LGTM

PR #124, opened 2024-05-03
UI-9384 - Preserve KPI titles when migrated KPI widgets from 4.3 to 5.0

Nouzbe:
This ensures that KPI titles manually entered by users on Atoti UI 4.3 are still honored when they open their dashboards after a 4.3 -> 5.0 content migration.

image

The changes might seem bigger than expected. It is because both Atoti UI 4.3 and 5.0 use the unique names of its members in order to identify a tuple, but they don't do it in the same order unfortunately.

Atoti UI 5.0 does:
- fields on columns (including measures if present)
- fields on rows (including measures if present)

Atoti UI 4.3 does:
- fields on rows (including measures if present)
- fields on columns (including measures if present)

### DIFF
--- a/src/4.3_to_5.0/getMigratedKpiTitles.ts
+++ b/src/4.3_to_5.0/getMigratedKpiTitles.ts
@@ -5,7 +5,13 @@ import {
   MdxUnknownCompoundIdentifier,
   parse,
 } from "@activeviam/activeui-sdk-5.0";
-import { getSpecificCompoundIdentifier, quote } from "@activeviam/mdx-5.0";
+import {
+  getSpecificCompoundIdentifier,
+  quote,
+  getMeasuresAxisName,
+  getMeasuresPositionOnAxis,
+  MdxSelect,
+} from "@activeviam/mdx-5.0";
 
 interface LegacyKpiTitle {
   title: string;
@@ -68,24 +74,43 @@ export function getMigratedKpiTitles(
   // Legacy widget states are not typed.
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
   legacyKpiState: any,
-  { mapping, cube }: { mapping: DataVisualizationWidgetMapping; cube: Cube },
+  {
+    legacyMdx,
+    mapping,
+    cube,
+  }: {
+    legacyMdx: MdxSelect;
+    mapping: DataVisualizationWidgetMapping;
+    cube: Cube;
+  },
 ): KpiWidgetState["titles"] {
   const legacyTitles = _getLegacyKpiTitles(legacyKpiState, { cube });
   const migratedTitles: KpiWidgetState["titles"] = {};
 
-  const memberUniqueNames: string[] = [];
+  const numberOfColumnFields = mapping.columns?.length ?? 0;
+
+  const measuresAxisName = getMeasuresAxisName(legacyMdx);
+  const measuresAxis = legacyMdx.axes.find(
+    (axis) => axis.name === measuresAxisName,
+  );
+  let measuresPositionInTuple = -1;
+  if (measuresAxis) {
+    const measuresPositionOnAxis = getMeasuresPositionOnAxis(measuresAxis);
+    measuresPositionInTuple = ["0", "COLUMNS"].includes(measuresAxisName)
+      ? measuresPositionOnAxis
+      : numberOfColumnFields + measuresPositionOnAxis;
+  }
+
+  // Atoti UI 5.0 KPI widgets expect tuple keys to express members in the following order:
+  // - column fields first
+  // - then row fields
+  const ordinalFields = [...(mapping.columns ?? []), ...(mapping.rows ?? [])];
+
   legacyTitles.forEach(({ title, tuple }) => {
-    // Atoti UI 5.0 KPI widgets expect tuple keys to express members in the following order:
-    // - column fields first
-    // - then row fields
-    const fields = [...(mapping.columns ?? []), ...(mapping.rows ?? [])];
-    fields.forEach((field) => {
-      if (field.type === "allMeasures") {
-        const measureName = tuple[`[Measures].[Measures]`]?.[0];
-        if (measureName) {
-          memberUniqueNames.push(`[Measures].[${measureName}]`);
-        }
-      } else if (field.type === "hierarchy") {
+    const memberUniqueNames: string[] = [];
+
+    ordinalFields.forEach((field) => {
+      if (field.type === "hierarchy") {
         const hierarchyUniqueName = quote(
           field.dimensionName,
           field.hierarchyName,
@@ -98,6 +123,16 @@ export function getMigratedKpiTitles(
         }
       }
     });
+
+    if (measuresPositionInTuple > -1) {
+      const measureName = tuple[quote("Measures", "Measures")][0];
+      memberUniqueNames.splice(
+        measuresPositionInTuple,
+        0,
+        quote("Measures", measureName),
+      );
+    }
+
     const tupleKey = memberUniqueNames.join(",");
     migratedTitles[tupleKey] = title;
   });
--- a/src/4.3_to_5.0/getMigratedKpiTitles.ts
+++ b/src/4.3_to_5.0/getMigratedKpiTitles.ts
@@ -48,9 +48,11 @@ function _getLegacyKpiTitles(
           specificCompoundIdentifier.measureName,
         ];
       } else if (specificCompoundIdentifier.type === "member") {
-        tuple[
-          `[${specificCompoundIdentifier.dimensionName}].[${specificCompoundIdentifier.hierarchyName}]`
-        ] = specificCompoundIdentifier.path;
+        const hierarchyUniqueName = quote(
+          specificCompoundIdentifier.dimensionName,
+          specificCompoundIdentifier.hierarchyName,
+        );
+        tuple[hierarchyUniqueName] = specificCompoundIdentifier.path;
       }
     }
     legacyTitles.push({ title, tuple });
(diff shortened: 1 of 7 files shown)


src/4.3_to_5.0/getMigratedKpiTitles.ts
```diff
@@ -0,0 +1,101 @@
...
+        compoundIdentifier,
+        { cube },
+      );
+      if (specificCompoundIdentifier.type === "measure") {
+        tuple[`[Measures].[Measures]`] = [
+          specificCompoundIdentifier.measureName,
+        ];
+      } else if (specificCompoundIdentifier.type === "member") {
+        tuple[
+          `[${specificCompoundIdentifier.dimensionName}].[${specificCompoundIdentifier.hierarchyName}]`
```
antoinetissier, line 52: You could use our `quote` function

src/4.3_to_5.0/getMigratedKpiTitles.ts
```diff
@@ -0,0 +1,101 @@
...
+    // - then row fields
+    const fields = [...(mapping.columns ?? []), ...(mapping.rows ?? [])];
+    fields.forEach((field) => {
+      if (field.type === "allMeasures") {
+        const measureName = tuple[`[Measures].[Measures]`]?.[0];
+        if (measureName) {
+          memberUniqueNames.push(`[Measures].[${measureName}]`);
+        }
+      } else if (field.type === "hierarchy") {
+        const hierarchyUniqueName = `[${field.dimensionName}].[${field.hierarchyName}]`;
```
antoinetissier, line 87: Same here

src/4.3_to_5.0/migrateKpi.ts
```diff
@@ -206,6 +213,29 @@ export function migrateKpi(
...
+    if (migratedTitles && Object.keys(migratedTitles).length > 0) {
+      migratedWidgetState.titles = migratedTitles;
+    }
+  } catch (error) {
+    // Migrating the KPI titles is a best effort.
+    // The migration script should not fail if this part errors.
+    console.warn(
+      `Could not migrate the titles of the featured values widget named "${legacyKpiState.name}". Underlying error:\n`,
+      error,
+    );
```
antoinetissier, line 230: It would be nice to add this to the error report instead of console.logging IMO

src/4.3_to_5.0/migrateKpi.ts
```diff
@@ -189,7 +190,13 @@ export function migrateKpi(
   const mapping = deriveMappingFromMdx({
     mdx: query.mdx,
     cube,
-    widgetPlugin: pluginWidgetKpi,
+    widgetPlugin: {
+      ...pluginWidgetKpi,
+      // This makes the "allMeasures" tile part of the generated mapping.
+      // This tile indicates the position of the measures on the queries axes, necessary to migrate KPI titles.
+      // It is then removed from the mapping, as Atoti UI 5.0 does not expect it.
```
antoinetissier, line 197: I'm not sure I understand: can't we just recognize for each post-widget-state-migration mapping attribute whether there are measures mapped to it ? And then migrate the titles based on that ?

antoinetissier (commented):
Just a few questions but it's looking good !

src/4.3_to_5.0/migrateKpi.ts
Nouzbe: I agree. I initially did not see how to do it without refactoring, but it seems that I can do it using a `PartialMigrationError`.

src/4.3_to_5.0/migrateKpi.ts
Nouzbe: The problem is the position of `ALL_MEASURES` relative to level fields. Indeed, the order of the measure or level identifiers in the tuple matters for KPI titles to work in Atoti UI 5. They have to correspond to their order on the cellset axes.

 But since `pluginWidgetKpi` has `doesSupportMeasuresRedirection: false`, `deriveMappingFromMdx` does not include the `ALL_MEASURES` tile. This prevents us from knowing where the measure unique name should be placed in the tuple.

The simplest solution I found to that problem is to derive the mapping with `doesSupportMeasuresRedirection: true` in order to receive the `ALL_MEASURES` tile and then to filter it out from the mapping before returning the migrated widget state.

antoinetissier: Thanks for the explanation, it's simple so that sounds good to me if you want to keep it like this.

Just an alternative, we could avoid this trick while keeping things simple IMO:
```
const measuresAxisName = getMeasuresAxisName(mdx);
const positionOfAllMeasuresWithinAxis = getMeasuresPositionOnAxis(
   mdx.axes.find((axis) => axis.name === measuresAxisName)!,
);
```

This is what we do in [`deriveGenericMappingFromMdx`](https://github.com/activeviam/atoti-ui/blob/95e3ec5c9b95775eb5b6a56f465facd38b937d11/packages/widget/src/deriveGenericMappingFromMdx.ts#L152-L155).

antoinetissier (approved):

Nouzbe:
@antoinetissier can you have a look at the last commit and make sure this is what you had in mind?

src/4.3_to_5.0/getMigratedKpiTitles.ts
```diff
@@ -0,0 +1,106 @@
...
+  const migratedTitles: KpiWidgetState["titles"] = {};
+
+  const memberUniqueNames: string[] = [];
+  legacyTitles.forEach(({ title, tuple }) => {
+    // Atoti UI 5.0 KPI widgets expect tuple keys to express members in the following order:
+    // - column fields first
+    // - then row fields
+    const fields = [...(mapping.columns ?? []), ...(mapping.rows ?? [])];
+    fields.forEach((field) => {
+      if (field.type === "allMeasures") {
```
antoinetissier, line 83: Instead of relying on the field type, couldn't you insert measures at `measuresPositionOnAxis` if it is defined ?

antoinetissier (commented):
Thanks for assigning back yo me.

Maybe I'm missing something, but I was imagining something like the following, where we compute the measures position on the axis once and for all, and insert at the right position based on it for each tuple.

```
export function getMigratedKpiTitles(
  // Legacy widget states are not typed.
  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
  legacyKpiState: any,
  { mapping, cube }: { mapping: DataVisualizationWidgetMapping; cube: Cube },
): KpiWidgetState["titles"] {
  const legacyTitles = _getLegacyKpiTitles(legacyKpiState, { cube });
  const migratedTitles: KpiWidgetState["titles"] = {};

  const numberOfColumnFields = mapping.columns?.length ?? 0;

  const mdx = legactWidgetState.query.mdx;
  const measuresAxisName = getMeasuresAxisName(mdx);
  const measuresAxis = mdx.axes.find((axis) => axis.name === measuresAxisName);
  let measuresPositionInTuple = -1;
  if (measuresAxis) {
    const measuresPositionOnAxis = getMeasuresPositionOnAxis(measuresAxis);
    measuresPositionInTuple =
      measuresAxisName === "columns"
        ? measuresPositionOnAxis
        : numberOfColumnFields + measuresPositionOnAxis;
  }

  // Atoti UI 5.0 KPI widgets expect tuple keys to express members in the following order:
  // - column fields first
  // - then row fields
  const ordinalFields = [...(mapping.columns ?? []), ...(mapping.rows ?? [])];

  legacyTitles.forEach(({ title, tuple }) => {
    const memberUniqueNames: string[] = [];

    ordinalFields.forEach((field) => {
      if (field.type === "hierarchy") {
        const hierarchyUniqueName = quote(
          field.dimensionName,
          field.hierarchyName,
        );
        const namePath = tuple[hierarchyUniqueName];
        if (namePath) {
          memberUniqueNames.push(
            `${hierarchyUniqueName}.${quote(...namePath)}`,
          );
        }
      }
    });

    if (measuresPositionInTuple > -1) {
      const measureName = tuple[quote("Measures", "Measures")];
      memberUniqueNames.splice(
        measuresPositionInTuple,
        0,
        quote("Measures", measureName),
      );
    }

    const tupleKey = memberUniqueNames.join(",");
    migratedTitles[tupleKey] = title;
  });

  return migratedTitles;
}
```

antoinetissier (commented):
Looks great !
Re-reading the code, I am just noticing this https://github.com/activeviam/atoti-ui-migration/pull/124/commits/999513ebc157117ac5d89cca1f22ff0b06fd0a42#diff-a9ab507bbcfeb1b4251d5fded05f4930172c9f29fb65a73ab51246091dea1c53R118-R123
Can it actually happen that a hierarchy is not expressed in the tuple (i.e. that `namePath` is `undefined`) ?
- If it should not legitimately happen (which I think is the case) I'd rather be explicit and throw a clear error if it does.
- If it can legitimately happen, then I don't think my proposition works, because then the index of measures within the tuple can vary.

src/4.3_to_5.0/migrateKpi.ts
```diff
@@ -206,6 +210,25 @@ export function migrateKpi(
...
+      });
+      migratedWidgetState.titles = migratedTitles;
+    }
+  } catch (error) {
+    // Migrating the KPI titles is a best effort.
+    // The migration script should not fail if this part errors.
+    throw new PartialMigrationError(
+      `Could not migrate the titles of the featured values widget named "${legacyKpiState.name}"`,
+      serializeWidgetState(migratedWidgetState),
+    );
```
antoinetissier, line 229: I think it'd be helpful to preserve the error message when wrapping it in `PartialMigrationError`, to help consumers of the CLI fix that error.

Nouzbe:
> Can it actually happen that a hierarchy is not expressed in the tuple (i.e. that namePath is undefined) ?

It can legitimately happen:
- add Currency and create a title
- remove Currency
- add City and create a title

=> There is a tuple in `legacyTitles` which does not have any entry for `[Geography].[City]`. If we throw an error here in this case, it means that we fail to migrate the KPI titles for this widget. So I think that we shouldn't throw here.

I also think that your suggestion still works in this case.

antoinetissier:
Right, with this in mind we shouldn't throw, although then cleaning up the obsolete titles would be better than not doing anything IMO.

Nouzbe:
I'll merge this: I consider the cleanup as out of scope here. I think that it would be nice to have it as a part of Admin UI, rather than coupled to the migration script. For instance, we could have a "Clean up" button that does an audit of the content and offers clean up actions to the user, like removing stale references to tuples from widget states or deleting saved dashboards/widgets/filters that do not have an entry under `structure`.

PR #125, opened 2024-05-14
UI-9409 - Fix Error: Cannot perform 'get' on a proxy that has been revoked when attempting to migrate while removing widgets

NZepeda:
## Summary

The issue is described in this ticket: https://activeviam.atlassian.net/browse/UI-9409.

### Problem
The problem is that when attempting to run the migration script, the following error is thrown: `"TypeError: Cannot perform 'get' on a proxy that has been revoked"`. This is caused by `immer`'s `produce` function. Unfortunately, I'm not `100%` why this error is being thrown in the first place. The fix I went for was to remove that `produce` function, extract the logic into a function and add a test for it.


antoinetissier (commented):
@NZepeda there seems to be an install/build error in the CI.
Can you look into it ?

antoinetissier:
@NZepeda have you been able to investigate more ?

NZepeda:
@antoinetissier Yes, the issue seems to be this double usage of the `produce` function. It seems to be causing a "proxy leak" where the returned `dashboardWithWidgetsRemoved` object still contains a property that is a `Proxy`. This causes the error we're seeing `TypeError: Cannot perform 'get' on a proxy that has been revoked`. Like I did on this PR, removing the usage of the `produce` function here fixes the issue. I actually created a dummy [PR](https://github.com/activeviam/atoti-ui/pull/4758) in the Atoti-UI repo where I added two tests, one using the `produce` function and the other not and I'm able to reproduce the error. The test using the `produce` function throws the same error while the other passes as expected. I haven't had the chance to pinpoint exactly _**why**_ this is happening yet.

PR #127, opened 2024-05-16
UI-8797 - Migrate smart filtering settings

antoinetissier:
Fixes [UI-8797](https://activeviam.atlassian.net/browse/UI-8797)

4.3 settings documentation: https://docs.activeviam.com/products/atoti/ui/4.3/dev/reference/settings.html#memberselectionsmartfiltering
4.3 UI: https://activepivot-ranch.activeviam.com:5900/ui/index.html
Corresponding content server: https://activepivot-ranch.activeviam.com:5900/content/ui/index.html

Note: some 4.3 settings don't have an equivalent in 5.x.
Most importantly: `memberSelection.smartFiltering.includeSlicingHierarchy`, which was an opt-in in 4.3, whereas all slicing hierarchies are included (not configurable) in 5.x.

Also, smart filtering was configurable per cube in 4.3, but not in 5.x.
I chose to ignore cube-specific settings, as you would use them only if you have more than one cube, in which case we don't know which one to take for 5.x anyway.
Completely out of scope, but note that we did have an internal request for per-cube settings in 5.x as well: https://activeviam.atlassian.net/browse/UI-8588


src/4.3_to_5.0/migrateSettingsFolder.ts
```diff
@@ -76,6 +78,44 @@ function migrateSettingsMap(legacySettingsMap: {
     );
   }
 
+  /* Smart filtering settings start. */
```
Nouzbe, line 81: To split code, it would be more robust to do:

```
/**
 * Adds the smart-filtering settings from `legacySettingsMap` into `migratedSettingsMap`.
 * Mutates `migratedSettingsMap`.
 */
migrateSmartFilteringSettings(migratedSettingsMap, legacySettingsMap);
```
rather than to add these comments, which are likely to become out of place in the future.

Another benefit is that this function would be easily unit testable, with unit tests that will be more maintainable than the one that generates a 3K lines snapshot.

src/4.3_to_5.0/migrateSettingsFolder.ts
```diff
@@ -76,6 +78,44 @@ function migrateSettingsMap(legacySettingsMap: {
...
+    migratedSettingsMap["smartFiltering.isEnabled"] = false;
+  } else {
+    const nonEmptyEvaluationMeasureUniqueName =
+      legacySettingsMap["memberSelection.smartFiltering.nonEmptyMeasure"];
+    if (nonEmptyEvaluationMeasureUniqueName !== undefined) {
+      const [, nonEmptyEvaluationMeasureName] = unquote(
+        nonEmptyEvaluationMeasureUniqueName,
+      );
+      // @ts-expect-error TypeScript rightfully complains that this setting key does not exist in 5.0
+      // It was added only in 5.1 indeed, but it makes sense to migrate it as an intermediate step.
```
Nouzbe, line 94: Why does it make sense? Wouldn't the fact that the migration script yields settings that do nothing be seen as a bug by customers?

Why not do it in `5.0_to_5.1/migrateSettingsFolder`?

antoinetissier: The `nonEmptyEvaluationMeasureName` setting exists in 4.3 and 5.1, but not in 5.0.
You do want to preserve it if you go from 4.3 to 5.1.

Implementation-wise, right now the migration script works only with iterative steps (4.3 to 5.0, 5.0 to 5.1), etc.
So if you discard the 4.3 setting while going from 4.3 to 5.0, you would not end up having it in the 5.1 output.

**Alternative 1**
Indeed do not discard it when going from 4.3 to 5.0, but actually migrate it to its new syntax only in the 5.0 to 5.1 step. But there would be a type error as well, as the 4.3 key does not exist in the 5.0 settings.

**Alternative 2**
Do an extra step outside of the iterative steps to avoid the type error.

But:
- We would start doing things like `if fromVersion === and toVersion === then....` in `migrateContentServer` which was supposed to be generic. This is a bit what I'm doing for calculated measures though.
- We would explode the logic of migrate settings in several places.

No solution is ideal theoretically, but pragmatically the current state of the PR seemed like the best option to me.

Nouzbe: I agree that Alternative 2 is bad.

But I picture a customer asking why the setting that we generated during their migation to 5.0 does not work in 5.0 and I wouldn't be able to tell them _"yes it's on purpose, it's because of how our migration script is made internally and I did not want to add a ts-expect-error"_. In practice, I would have to agree that it is a bug and we would have to fix it (=> to backport the 5.1 setting to 5.0).

I still remember the feedback from SCB at the London connect with their 2 main issues:
- upgrades are two costly
- the product is so frequently bugged that they feel like we don't do QA at all and that we consider our customers as our QA testers

I generally don't want to introduce something that I consider right off the bat, and this kind of face-to-face feedback probably makes it even more important.

So I like Alternative 1 even if it means some `ts-expect-error`.

Or I can think of an Alternative 3, where the step 4.3 -> 5.0 transforms the setting key into something that makes it clear that it does not work in 5.0 and will work in 5.1. For instance `__upgrade_to_5.1_to_use_this__<setting_key>`. And then make the 5.0 -> 5.1 step migrate it again to what it needs.

Finally Alternative 4 which might be the best: actually backport the setting right now so that this problem does not exist.

What do you think?

Nouzbe (approved):
Looks good 👍

PR #128, opened 2024-07-29
UI-9526 - Ensures that the `behaviorOnError` flag is respected

NZepeda:
## Summary

This PR ensures that when a 4.3 widget is being migrated if it throws an error, it is kept as is and that the remaining widgets are still migrated as expected.

## Testing 
All tests should pass.

### DIFF
--- a/src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
+++ b/src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
@@ -1336,7 +1336,7 @@ exports[`migrate_43_to_50 returns a valid ActiveUI5 /ui folder on a real life in
               "entry": {
                 "canRead": true,
                 "canWrite": true,
-                "content": "{}",
+                "content": "{"name":"Page filters","type":"container","value":{"style":{},"showTitleBar":true,"body":{},"containerKey":"filters"}}",
                 "isDirectory": false,
                 "lastEditor": "admin",
                 "owners": [
@@ -1608,7 +1608,7 @@ exports[`migrate_43_to_50 returns a valid ActiveUI5 /ui folder on a real life in
                   "entry": {
                     "canRead": true,
                     "canWrite": true,
-                    "content": "{"name":"Page filters","widgetKey":"filters"}",
+                    "content": "{"name":"Page filters"}",
                     "isDirectory": true,
                     "lastEditor": "admin",
                     "owners": [
--- a/src/4.3_to_5.0/__test_resources__/smallLegacyUIFolderWithInvalidWidget.ts
+++ b/src/4.3_to_5.0/__test_resources__/smallLegacyUIFolderWithInvalidWidget.ts
@@ -1,0 +1,162 @@
+/**
+ * The shortened version of the content of the /ui folder on a Content Server, useful for unit tests.
+ * Contains an invalid widget whose `containerKey` is not valid.
+ */
+export const smallLegacyUIFolderWithInvalidWidget = {
+  entry: {
+    isDirectory: true,
+    owners: ["admin"],
+    readers: ["admin"],
+    timestamp: 1607879725132,
+    lastEditor: "admin",
+    canRead: true,
+    canWrite: true,
+  },
+  children: {
+    bookmarks: {
+      entry: {
+        isDirectory: true,
+        owners: ["ROLE_CS_ROOT"],
+        readers: ["ROLE_USER"],
+        timestamp: 1607879735685,
+        lastEditor: "admin",
+        canRead: true,
+        canWrite: true,
+      },
+      children: {
+        content: {
+          entry: {
+            isDirectory: true,
+            owners: ["ROLE_USER"],
+            readers: ["ROLE_USER"],
+            timestamp: 1607879735685,
+            lastEditor: "admin",
+            canRead: true,
+            canWrite: true,
+          },
+          children: {
+            "158": {
+              entry: {
+                content:
+                  '{"description":"Widget with invalid container key","name":"Invalid widget","type":"container","value":{"style":{},"showTitleBar":false,"containerKey":"invalid-container-key","body":{"serverUrl":"","mdx":"SELECT NON EMPTY [Measures].[contributors.COUNT] ON COLUMNS FROM [EquityDerivativesCube] WHERE [Geography].[City].[ALL].[AllMember].[New York] CELL PROPERTIES VALUE, FORMATTED_VALUE, BACK_COLOR, FORE_COLOR, FONT_FLAGS","contextValues":{},"updateMode":"once","ranges":{}}}}',
+                isDirectory: false,
+                owners: ["admin"],
+                readers: ["admin"],
+                timestamp: 1607879735685,
+                lastEditor: "admin",
+                canRead: true,
+                canWrite: true,
+              },
+            },
+            "777": {
+              entry: {
+                content:
+                  '{"description": "Valid widget","name": "Valid widget","type": "container", "value": {"style": {},"showTitleBar": false,"containerKey": "pivot-table","body": {"serverUrl": "","mdx": "SELECT NON EMPTY [Measures].[contributors.COUNT] ON COLUMNS FROM [EquityDerivativesCube] WHERE [Geography].[City].[ALL].[AllMember].[New York] CELL PROPERTIES VALUE, FORMATTED_VALUE, BACK_COLOR, FORE_COLOR, FONT_FLAGS","contextValues": {},"updateMode": "once", "ranges": {}}}}',
+                isDirectory: false,
+                owners: ["admin"],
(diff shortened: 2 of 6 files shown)


src/getMigrateSavedWidgets.ts
```diff
@@ -68,7 +68,7 @@ export const getMigrateSavedWidgets =
 
     for (const fileId in content.children) {
       if (errorReport.widgets?.[fileId] && behaviorOnError !== "keep-going") {
-        return;
+        continue;
```
NZepeda, line 71: Previously, returning here would cause the entire script to "short-circuit" causing none of the remaining widgets to be migrated.

src/4.3_to_5.0/__test_resources__/smallLegacyUIFolderWithInvalidWidget.ts
```diff
@@ -0,0 +1,186 @@
+/**
+ * The shortened version of the content of the /ui folder on a Content Server, useful for unit tests.
+ * Contains a widget whose `containerKey` is not valid and another widget with a filter with an invalid hierarchy.
```
NZepeda, line 3:
```suggestion
 * Contains a widget whose `containerKey` is not valid and another widget with a filter on a hierarchy that does not exist in the test cube.
```

src/migrateContentServer.test.ts
```diff
@@ -232,4 +233,97 @@ describe("migrateContentServer", () => {
...
+
+    const savedWidgetContentBeforeMigration =
+      contentServerBeforeMigration.children?.ui.children?.bookmarks.children
+        ?.content.children?.["158"].entry.content;
+
+    const savedWidgetContentAfterMigration =
+      contentServer?.children?.ui?.children?.widgets?.children?.content
+        ?.children?.["158"].entry.content;
+
+    expect(savedWidgetContentBeforeMigration).toBe(
```
Nouzbe, line 275: Let's add some comments here:

```suggestion
    // The widget with id "158" has an invalid widget key and its migration will therefore fail.
    // Since `behaviorOnError` is set to "keep-original", the output content still contains the state of this saved widget exactly as it was before the migration.
    expect(savedWidgetContentBeforeMigration).toBe(
```

src/migrateContentServer.test.ts
```diff
@@ -232,4 +233,97 @@ describe("migrateContentServer", () => {
...
+      savedWidgetContentAfterMigration,
+    );
+  });
+
+  it("keeps 5.0 version of the widget, when there is an error when attempting to migrate the widget from 5.0 to 5.1 when the `behaviorOnError` flag is set to `keep-last-successful-version`.", async () => {
```
Nouzbe, line 280:
```suggestion
  it("keeps the 5.0 version of the widget, when migrating from 4.3 to 5.1 with `behaviorOnError` set to `keep-last-successful-version` and the 5.0 to 5.1 step fails", async () => {
```

src/migrateContentServer.test.ts
```diff
@@ -232,4 +233,97 @@ describe("migrateContentServer", () => {
...
+      contentServer?.children?.ui?.children?.widgets?.children?.content
+        ?.children?.["1231"].entry.content,
+    );
+
+    expect(savedWidgetContentAfterMigration.filters).toBeDefined();
+    expect(savedWidgetContentAfterMigration.filters).not.toHaveLength(0);
+
+    const migratedWidgetFilter = savedWidgetContentAfterMigration.filters[0];
+
+    // The filters in 5.0 are represented as strings.
```
Nouzbe, line 326:
```suggestion
    // The widget with id "1231" contains a filter referring to a hierarchy which does not exist in the cube.
    // This leads to an error in the 5.0 -> 5.1 step.
    // As `behaviorOnError` is set to "keep-last-successful-version", the output content contains the "5.0" version of this widget.
    // The filters in 5.0 are represented as strings.
```

src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -434,19 +434,6 @@ export async function migrate_43_to_50(
             });
             counters.widgets.success++;
           } catch (error) {
-            if (error instanceof PartialMigrationError) {
```
Nouzbe, line 437: Does it make a difference to move this block after the call to `_addErrorToReport`? I don't mind, but just trying to understand the change.

src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
```diff
@@ -1336,7 +1336,7 @@ exports[`migrate_43_to_50 returns a valid ActiveUI5 /ui folder on a real life in
               "entry": {
                 "canRead": true,
                 "canWrite": true,
-                "content": "{}",
+                "content": "{"name":"Page filters","type":"container","value":{"style":{},"showTitleBar":true,"body":{},"containerKey":"filters"}}",
```
Nouzbe, line 1339: Do you understand this change?

src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
```diff
@@ -1608,7 +1608,7 @@ exports[`migrate_43_to_50 returns a valid ActiveUI5 /ui folder on a real life in
                   "entry": {
                     "canRead": true,
                     "canWrite": true,
-                    "content": "{"name":"Page filters","widgetKey":"filters"}",
+                    "content": "{"name":"Page filters"}",
```
Nouzbe, line 1611: Same question here.

src/4.3_to_5.0/migrate_43_to_50.ts
```diff
@@ -458,6 +445,25 @@ export async function migrate_43_to_50(
...
             });
+            if (error instanceof PartialMigrationError) {
+              counters.widgets.partial++;
+              migratedWidget = error.migratedWidgetState;
+            } else {
+              counters.widgets.failed++;
+              // The migration failed.
+              // The widget is kept as is.
+              migratedWidget = bookmark;
+              widgets[fileId] = migratedWidget;
```
Nouzbe, line 456: Why do we need to do this here? Isn't it already taken care of in https://github.com/activeviam/atoti-ui-migration/blob/c88337116110b275c174540fe940b69d0bfcec9d/src/4.3_to_5.0/migrate_43_to_50.ts#L464?

Nouzbe (commented):
I have a few questions:

src/4.3_to_5.0/migrate_43_to_50.ts
NZepeda: @Nouzbe I undid my changes within `migrate_4.3_to_5.0` as indeed, preserving the bookmarks content was handled by this. There was a slight modification to one of the tests but all tests should still pass. This makes this PR more simple 👍

src/migrateContentServer.test.ts
```diff
@@ -232,4 +233,107 @@ describe("migrateContentServer", () => {
...
+      fromVersion: "4.3",
+      toVersion: "5.1",
+      keysOfWidgetPluginsToRemove: [],
+      doesReportIncludeStacks: false,
+      shouldUpdateFiltersMdx: true,
+      behaviorOnError: "keep-original",
+    });
+
+    // In 4.3 the content of the widget is kept in the `body` which is within the bookmark's `content` string.
+    // If the migration of the widget fails, we keep the content of the widget intact but move its location in order to maintain the structure expected in 5.0/5.1
```
NZepeda, line 268:
```suggestion
    // If the migration of the widget fails, we keep the content of the widget intact but move its location in order to maintain the structure expected in 5.0/5.1.
```

NZepeda:
@Nouzbe I'm passing this back to you. In my latest commit, I undid the changes to `migrate_43_to_50` because as you pointed out, the preservation of the original bookmark content was already being handled. The difference with my previous changes, was that I was keeping the entire 4.3 bookmarks `value` which looks like the following:

```
{
  "value": {
    "style": {},
    "showTitleBar": false,
    "containerKey": "invalid-container-key",
    "body": {
      "serverUrl": "",
      "mdx": "SELECT NON EMPTY [Measures].[contributors.COUNT] ON COLUMNS FROM [EquityDerivativesCube] WHERE [Geography].[City].[ALL].[AllMember].[New York] CELL PROPERTIES VALUE, FORMATTED_VALUE, BACK_COLOR, FORE_COLOR, FONT_FLAGS",
      "contextValues": {},
      "updateMode": "once",
      "ranges": {}
    }
  }
}
```

The existing changes were already valid though, just keeping the `value.body` portion of the bookmark. So I reverted my changes which allow the `migrate_43_to_50.ts` file to remain untouched and thus simplify this PR a bit more. Now the changes mostly revolve around the tests ensuring that the `behaviorOnError` flag is respected. 

Please let me know if you have any other questions or concerns, I'm happy to address them.

PR #129, opened 2024-07-29
UI-9566 - Ensure a best effort migration of the KPI titles even when some titles could not be successfully migrated.

NZepeda:
### Issue
As the [ticket](https://activeviam.atlassian.net/browse/UI-9566) states, when migrating KPIs containing custom titles, if the migration of one of the titles throws an error, none of the custom titles are migrated.

On this specific ticket, the issue was that when the KPI contained a title with an empty `tupleKey` as seen below, the process would error out resulting in none of the titles (even the valid ones) being migrated. 

```
"": {
   title: "Title with empty tupleKey"
}
```

### Solution
This PR aims to improve on that by ensuring that even if there is an error, the successfully migrated titles are persisted and the `KPI` widget is marked as `Partially migrated`.

Corresponding ticket: https://activeviam.atlassian.net/browse/UI-9566

### Reviewing
Ensure the new and updated tests make sense


src/4.3_to_5.0/getMigratedKpiTitles.ts
```diff
@@ -125,7 +132,8 @@ export function getMigratedKpiTitles(
     });
 
     if (measuresPositionInTuple > -1) {
-      const measureName = tuple[quote("Measures", "Measures")][0];
+      const measureName = tuple[quote("Measures", "Measures")]?.[0];
+
```
NZepeda, line 136:
```suggestion
      const measureName = tuple[quote("Measures", "Measures")][0];
```

src/4.3_to_5.0/__test_resources__/legacyKpi.ts
```diff
@@ -31,6 +31,9 @@ export const legacyKpi: LegacyWidgetState = {
       configuration: {
         featuredValues: {
           locations: {
+            "": {
+              title: "Empty measure title",
+            },
```
NZepeda, line 36: For this fix, I went with dropping titles that have an empty tuple key and ensuring the rest of the titles are still migrated. However, I'm not 100% sure how one can even land in this state. It doesn't seem like it's valid so I dropped it, but if it is better to ensure this kind of custom title is supported, I can adapt the change to ensure it's also migrated.

src/migrateContentServer.test.ts
```diff
@@ -232,4 +232,42 @@ describe("migrateContentServer", () => {
       undefined,
     );
   });
+
+  it("migrates the kpi custom titles when the original version contains a custom title with a tupleKey that is empty.", async () => {
```
NZepeda, line 236:
```suggestion
  it("migrates the KPI custom titles while dropping custom titles with an empty `tupleKey` for their location", async () => {
```

src/migrateContentServer.test.ts
```diff
@@ -232,4 +232,42 @@ describe("migrateContentServer", () => {
...
+      doesReportIncludeStacks: false,
+      shouldUpdateFiltersMdx: true,
+      behaviorOnError: "keep-original",
+    });
+
+    const migratedKpiContent = JSON.parse(
+      contentServer.children?.ui.children?.widgets.children?.content.children?.[
+        "kpi"
+      ].entry.content,
+    );
```
NZepeda, line 265:
```suggestion
    // The widget with id "kpi" contains a custom title referenced with an empty `tupleKey`.
    // Because this title doesn't represent any tuple, it is dropped.
    const kpiContentBeforeMigration = contentServer.children?.ui.children?.widgets.children?.content.children?.[
        "kpi"
      ].entry.content;
      
    const migratedKpiContent = JSON.parse(kpiContentBeforeMigration);
```

src/4.3_to_5.0/accumulateMigratedKpiTitles.ts
```diff
@@ -43,49 +43,55 @@ function _getLegacyKpiTitles(
...
+          tuple[`[Measures].[Measures]`] = [
+            specificCompoundIdentifier.measureName,
+          ];
+        } else if (specificCompoundIdentifier.type === "member") {
+          const hierarchyUniqueName = quote(
+            specificCompoundIdentifier.dimensionName,
+            specificCompoundIdentifier.hierarchyName,
+          );
+          tuple[hierarchyUniqueName] = specificCompoundIdentifier.path;
+        }
```
NZepeda, line 63: Looks like a big change but all I did was wrap this logic block in an `if` statement.

src/4.3_to_5.0/accumulateMigratedKpiTitles.ts
```diff
@@ -107,35 +113,44 @@ export function getMigratedKpiTitles(
...
-      );
-    }
 
-    const tupleKey = memberUniqueNames.join(",");
-    migratedTitles[tupleKey] = title;
+      const tupleKey = memberUniqueNames.join(",");
+      migratedTitles[tupleKey] = title;
+    } catch (error) {
+      errors.push(error as Error);
+    }
```
NZepeda, line 147: Again, looks like a big change but I just wrapped this logic block in a try-catch.

src/4.3_to_5.0/accumulateMigratedKpiTitles.ts
```diff
@@ -107,35 +113,44 @@ export function getMigratedKpiTitles(
...
   });
 
-  return migratedTitles;
+  if (errors.length > 0) {
+    throw new Error(
+      `One or more errors occurred while migrating the titles for the kpi named ${legacyKpiState.name}: ` +
+        errors.map((e) => e.message).join(", "),
+    );
+  }
```
NZepeda, line 155: This function "swallows" any error until all `legacyTitles` have been attempted to be migrated. This throws at the end to ensure that the kpi widget that is being migrated is marked as `partially migrated`.

src/migrateContentServer.test.ts
```diff
@@ -232,4 +233,68 @@ describe("migrateContentServer", () => {
       undefined,
     );
   });
+
+  it("migrates the KPI custom titles while dropping custom titles with an empty `tupleKey` for their location", async () => {
```
NZepeda, line 237:
```suggestion
  it("migrates the KPI custom titles while dropping custom titles that could not be successfully migrated", async () => {
```

src/4.3_to_5.0/__test_resources__/smallLegacyUIFolderWithInvalidKpiTitle.ts
```diff
@@ -0,0 +1,241 @@
+/**
+ * The shortened version of the content of the /ui folder on a Content Server, useful for unit tests.
+ * Contains a KPI widget containing a custom title with an empty `tupleKey` for its location.
+ */
+export const smallLegacyUIFolderWithInvalidKpiTitle = {
```
Nouzbe, line 5: It would be nice to factorize the common part between this and the test resource introduced in #128, in order to avoid adding redundant content to this repo.

For instance, we could have a `emptyLegacyUIFolder` test resource and a `addLegacyBookmarkToUIFolder` test resource function that takes a map of id to legacy bookmark as argument and returns the "small legacy UI folder with ..." object, without mutating `emptyLegacyFolder`.

Nouzbe (commented):
Could you please merge #128 into main, rebase this branch and take the following comment into consideration before reassigning?

NZepeda:
Hi , sorry this took so long for me to get back to you. I finally had the time to return to this PR. To give you a brief recap of where things left off, this PR aims to fix the issue where some customers using `4.3` were running into an issue when attempting to migrate their KPI widgets over to `5.0`. The issue is that some KPIs have a "custom title" with an empty key such as: 

```js
"": {
   title: "Title with empty tupleKey"
}
```

This caused the migration script to error resulting in **_none_** of the custom KPI titles being migrated. I've updated the code in this PR so that in case there is an error when migrating the KPI titles, we keep going with the migration of the other titles and only "throw" the error afterward to mark the widget as "partially" migrated. 

In the latest commits, I've taken your suggestions into account. Specifically this [one](https://github.com/activeviam/atoti-ui-migration/pull/129#discussion_r1699899921). I created the `legacyEmtpyUIFolder` test resource file and the `addLegacyBookmarkToUIFolder` test resource function to prevent duplication amongst the test resource files. Let me know your thoughts on the latest changes.

src/4.3_to_5.0/migrateKpi.ts
```diff
@@ -210,20 +210,27 @@ export function migrateKpi(
...
       });
       if (migratedTitles && Object.keys(migratedTitles).length > 0) {
         migratedWidgetState.titles = migratedTitles;
       }
     }
   } catch (error) {
     // Migrating the KPI titles is a best effort.
+    // If there is an error while migrating one of the titles, the successfully migrated titles are still persisted.
+    if (migratedTitles && Object.keys(migratedTitles).length > 0) {
+      migratedWidgetState.titles = migratedTitles;
```
Nouzbe, line 231: If we encounter an error, this means that we will save the already migrated KPI titles, but we will still lose all the ones that weren't yet reached in the iteration, right?

A more ideal behavior would be to even keep doing the iteration and keep accumulating and saving the rest of KPI titles.

src/4.3_to_5.0/addLegacyBookmarkToUIFolder.ts
```diff
@@ -0,0 +1,61 @@
+import { ContentRecord } from "@activeviam/activeui-sdk-5.1";
+import { produce } from "immer";
+
+/**
+ * Returns a new `legacyUIFolder` with the given `legacyBookmarksToAdd` added.
+ * Useful for creating legacyUIFolders for tests.
+ * Does not mutate `legacyUIFolder`.
+ */
+export function addLegacyBookmarkToUIFolder(
```
Nouzbe, line 9: Let's move this function to within `__test_resources__` as it will be used just to generate test resources.

src/4.3_to_5.0/__test_resources__/emptyLegacyUIFolder.ts
```diff
@@ -0,0 +1,214 @@
...
...\n          {\n            "columnName": "ProductQtyMultiplier",\n            "functionName": "Caption"\n          }\n        ]\n      }\n    },\n    "defaultPermissions": {\n      "owners": [\n        "admin"\n      ],\n      "readers": [\n        "admin"\n      ]\n    },\n    "project.content-editor.configuration": {\n      "autoSwitchToFieldsOnEmptyWidget": true,\n      "advancedModeDropdownHidden": false,\n      "hideFieldsControls": false,\n      "mdx-common": {\n        "regularModeTab": "filters",\n        "advancedModeEnabled": false,\n        "advancedModeTab": "mdx"\n      }\n    },\n    "user.contextValues": {\n      "EquityDerivativesCube": {},\n      "EquityDerivativesCubeDist": {}\n    },\n    "user.filters": {\n      "EquityDerivativesCube": ["[Geography].[City].[ALL].[AllMember].[Berlin]"],\n      "EquityDerivativesCubeDist": []\n    },\n    "filtering.hierarchyToFilterType.[Geography].[City]": "explicit",\n    "filtering.hierarchyToFilterType.[Currency].[Currency]": "explicit",\n    "filtering.hierarchyToFilterType.[Trades].[Trades]": "topcount",\n    "servers.alias": {\n      "https://activepivot-ranch.activeviam.com:5900": "http://localhost:8080"\n    }\n  }\n}',
```
Nouzbe, line 185: I would prefer to not have this in `emptyLegacyUIFolder` and to have it actually be empty.

src/migrateContentServer.test.ts
```diff
@@ -18,6 +19,54 @@ jest.mock(`./4.3_to_5.0/generateId`, () => {
   };
 });
 
+/**
+ * Contains three legacy bookmark widgets.
+ * 1. Widget with an invalid container container key.
+ * 2. Widget with a filter on an invalid hierarchy.
+ * 3. A valid widget.
+ */
+const bookmarkContentWithInvalidWidgets = {
```
Nouzbe, line 28: It would be nice to move this to its own test resource (e.g. `invalidLegacyWidgets.ts`).

src/migrateContentServer.test.ts
```diff
@@ -285,9 +339,14 @@ describe("migrateContentServer", () => {
   });
 
   it("keeps the 5.0 version of the widget, when migrating from 4.3 to 5.1 with `behaviorOnError` set to `keep-last-successful-version` and the 5.0 to 5.1 step fails", async () => {
+    const legacyUIFolderWithInvalidWidgets = addLegacyBookmarkToUIFolder(
```
Nouzbe, line 342: I would also find it nice to have a dedicated test resource (e.g. called `legacyUIFolderWithInvalidWidgets.ts`), exporting this. My rationale is that future us are more likely to notice that this already exists in case we need to use it in another test. We will be less likely to recreate the same test resource multiple times.

src/migrateContentServer.test.ts
```diff
@@ -336,4 +395,87 @@ describe("migrateContentServer", () => {
     // The filters in 5.0 are represented as strings.
     expect(typeof migratedWidgetFilter.mdx).toBe("string");
   });
+
+  it("migrates the KPI custom titles while dropping custom titles that could not be successfully migrated", async () => {
+    const legacyUIFolderWithInvalidKpiTitle = addLegacyBookmarkToUIFolder(
```
Nouzbe, line 400: Here too, it would be nice to have dedicated test resources for:
- `legacyKPIWithInvalidTitle`
- `legacyUIFolderWithKpiWithInvalidTitle`

src/migrateContentServer.test.ts
```diff
@@ -336,4 +395,87 @@ describe("migrateContentServer", () => {
...
+        ui: legacyUIFolderWithInvalidKpiTitle,
+        pivot: smallLegacyPivotFolder,
+      },
+      entry: {
+        owners: [],
+        readers: [],
+        isDirectory: true,
+        canRead: true,
+        canWrite: false,
+        lastEditor: "Freddie Mercury",
```
Nouzbe, line 430: 👨‍🎤

src/migrateContentServer.test.ts
```diff
@@ -336,4 +395,87 @@ describe("migrateContentServer", () => {
...
+        timestamp: 0xbeef,
```
Nouzbe, line 431: 🥩

src/4.3_to_5.0/addLegacyBookmarkToUIFolder.ts
Nouzbe, line 9: This could take a `contentType` argument in order to be able to add saved dashboards, but also widgets or filters into the legacy UI folder.

Nouzbe (commented):
It would be great to use our new `addLegacyBookmarkToUIFolder` in order to generate our existing:
- `smallLegacyUIFolder`
- `smallLegacyUIFolderWithInvalidDashboard`
- `smallLegacyUIFolderWithInvalidFilter`

It would still be nice to have them as their own test resources, but this way we would cut down some duplicate content and lower the entropy of this repo.

src/migrateContentServer.test.ts
NZepeda: 🎸

src/4.3_to_5.0/addLegacyBookmarkToUIFolder.ts
NZepeda: This is for legacy "bookmarks" and back in the day we used to plop everything into the `bookmarks` folder, so no need to differentiate 👀

src/migrateContentServer.test.ts
NZepeda: Done, this got moved to `__test_resources__/legacyUIFolderWithInvalidWidgets.ts` ✅

src/migrateContentServer.test.ts
NZepeda: Done, I split this out to `__test_resources__/invalidLegacyWidgets.ts`

src/migrateContentServer.test.ts
NZepeda: Done, these got split out into `__test_resources__/legacyKpiWithInvalidTitle.ts` and `__test_resources__/legacyUIFolderWithKpiWithInvalidTitle.ts`

src/4.3_to_5.0/addLegacyBookmarkToUIFolder.ts
NZepeda: Done ✅

NZepeda:
> It would be great to use our new addLegacyBookmarkToUIFolder in order to generate our existing:
> * smallLegacyUIFolder
> * smallLegacyUIFolderWithInvalidDashboard
> * smallLegacyUIFolderWithInvalidFilter

> It would still be nice to have them as their own test resources, but this way we would cut down some duplicate content and lower the entropy of this repo.

This is done via one of the latest commits 🫡

src/4.3_to_5.0/migrateKpi.ts
NZepeda: Actually, the behavior is already as you suggest, internally `accumulatedMigratedKpiTitles` "swallows" any error until all the titles have attempted to be migrated. Once all titles have attempted to be migrated, it throws an error to mark the widget "partially" migrated.

NZepeda:
@Nouzbe Your latest suggestions have been addressed in the latest commits 👍

src/4.3_to_5.0/__test_resources__/legacyUIFolderWithInvalidKpi.ts
```diff
@@ -0,0 +1,13 @@
+import { addLegacyBookmarkToUIFolder } from "./addLegacyBookmarkToUIFolder";
+import { emptyLegacyUIFolder } from "./emptyLegacyUIFolder";
+import { legacyKpiWithInvalidTitle } from "./legacyKpiWithInvalidTitle";
+
+/**
+ * Legacy UI folder containing a single kpi bookmark with an invalid kpi title
```
NZepeda, line 6:
```suggestion
 * A legacy UI folder containing a single KPI bookmark with an invalid KPI title.
```
```suggestion
 * Legacy UI folder containing a single kpi bookmark with an invalid kpi title
```

src/4.3_to_5.0/__test_resources__/legacyInvalidDashboardBookmark.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Legacy content entry representing a bookmark of an invalid dashboard.
```
NZepeda, line 2:
```suggestion
 * The content entry representing a bookmark of an invalid dashboard.
```
```suggestion
 * Legacy content entry representing a bookmark of an invalid dashboard.
```

src/4.3_to_5.0/__test_resources__/legacyInvalidFilterBookmark.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Content entry representing a legacy filter containing an invalid property
```
NZepeda, line 2:
```suggestion
 * The content entry representing a legacy filter containing an invalid property.
```

src/4.3_to_5.0/__test_resources__/legacyKpiWithInvalidTitle.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * The content entry for a legacy bookmark representing a KPI widget
```
NZepeda, line 2:
```suggestion
 * The content entry for a legacy bookmark representing a KPI widget.
```
```suggestion
 * The content entry for a legacy bookmark representing a KPI widget
```

src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
```diff
@@ -2404,86 +2397,6 @@ exports[`migrate_43_to_50 returns a valid ActiveUI5 /ui folder on a small input
...
-                  "admin",
-                ],
-                "readers": [
-                  "admin",
-                ],
-              },
-            },
-            "settings": {
-              "entry": {
-                "content": "{"theme":"dark-activeviam","search.maxResults":10,"userFilters.areEnabled":true,"drillthrough.defaultSelectedColumns":{"EquityDerivativesCube":["delta","gamma","pnlVega","Desk","Currency","Date","HostName"],"EquityDerivativesCubeDist":["BumpedMtmDown","ProductQtyMultiplier","vega","rho","productId","pnlVega","pnlDelta","pnl","gamma","delta","TradeId","ProductBaseMtm"]}}",
```
Nouzbe, line 2423: Oops, it looks like we're losing the settings conversion part of this test snapshot. It's probably because some settings that were part of a legacy UI folder test resource were lost in the process. Can you please make sure that we don't lose this?

src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
```diff
@@ -2247,12 +2247,8 @@ exports[`migrate_43_to_50 returns a valid ActiveUI5 /ui folder on a small input
                     "content": "{"name":"hidden grand totals"}",
                     "isDirectory": true,
                     "lastEditor": "admin",
-                    "owners": [
-                      "admin",
```
Nouzbe, line 2251: Here too, it would be nice to not lose this, in order to ensure that the migration script does not erase bookmark permissions.

src/4.3_to_5.0/migrateKpi.ts
Nouzbe: Got it now, thank you 👍

Nouzbe (commented):
The KPI titles change looks good. The test resources refactor looks good too. There's just a couple instances where we've lost some content in our snapshots (I assume because we've lost them in some test resources). See comments below.

Otherwise, good to go!

src/4.3_to_5.0/__test_resources__/emptyLegacyUIFolder.ts
NZepeda: @Nouzbe Removing this would mean losing the `settings` in the snapshot for the `migrate_43_to_50` test since now we're generating `smallLegacyUIFolder` with this file and injecting the bookmark content via the `addLegacyBookmarkToUIFolder`. I'll keep this as is.

src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
```diff
@@ -2397,6 +2404,86 @@ exports[`migrate_43_to_50 returns a valid ActiveUI5 /ui folder on a small input
       },
     },
     "users": {
+      "children": {
```
NZepeda, line 2407: This change results from re-adding the `settings` to the `emptyLegacyUIFolder`. With the changes in this commit, this file remains unchanged in this PR.

src/4.3_to_5.0/__snapshots__/migrate_43_to_50.test.ts.snap
NZepeda: I've added a new commit to avoid losing this information. However, it means re-adding the `settings` to the `emptyLegacyUIFolder` since we are now generating test resources such as `smallLegacyUIFolder` via the `emptyLegacyUIFolder` and `addLegacyBookmarkToUIFolder`.

NZepeda:
@Nouzbe With the latest commit, I've ensured that we are not losing any data in the `migrate_43_to_50` test snapshot. However, that meant re-adding the previously removed settings to the `emptyLegacyUIFolder`. This is required since test resources such as the `smallLegacyUIFolder` are being generated via the `emptyLegacyUIFolder` test resource and the `addLegacyBookmarkToFolder` function.

Nouzbe (approved):
A different approach could be to have a `addLegacySettings` function, which adds given settings to a legacy UI folder. This way, `smallLegacyUIFolder` could be generated from an actually empty `emptyLegacyUIFolder` using this function.

If you like this way better, I'll leave it up to you if you prefer to do it in this PR or separately.

This is good to go, thank you for bearing with me 👍

NZepeda:
> A different approach could be to have a `addLegacySettings` function, which adds given settings to a legacy UI folder. This way, `smallLegacyUIFolder` could be generated from an actually empty `emptyLegacyUIFolder` using this function.
> 
> If you like this way better, I'll leave it up to you if you prefer to do it in this PR or separately.
> 
> This is good to go, thank you for bearing with me 👍

I'll address this in a separate PR 👍

PR #130, opened 2024-09-24
UI-9786 - Explicitly preserve the default measure when migrating from 4.3 to 5.0

antoinetissier:
See UI-9786 and/or the multiline comment of `_addDefaultMeasureIfNoneIsExplicitlyExpressed` to understand the issue and the fix.

### DIFF
--- a/src/4.3_to_5.0/_addDefaultMeasureIfNoneIsExplicitlyExpressed.test.ts
+++ b/src/4.3_to_5.0/_addDefaultMeasureIfNoneIsExplicitlyExpressed.test.ts
@@ -1,0 +1,44 @@
+import { MdxSelect, parse, stringify } from "@activeviam/mdx-5.0";
+import { _addDefaultMeasureIfNoneIsExplicitlyExpressed } from "./_addDefaultMeasureIfNoneIsExplicitlyExpressed";
+import { dataModelsForTests } from "@activeviam/data-model-5.0";
+
+const cube = dataModelsForTests.sandbox.catalogs[0].cubes[0];
+
+describe("_addDefaultMeasureIfNoneIsExplicitlyExpressed", () => {
+  it("adds the default measure on columns when no measure is expressed on axes, but at least one hierarchy is", () => {
+    const mdx: MdxSelect = parse(`
+        SELECT
+         [Currency].[Currency].Members ON ROWS
+         FROM [EquityDerivativesCube]`);
+    expect(
+      stringify(_addDefaultMeasureIfNoneIsExplicitlyExpressed(mdx, { cube }), {
+        indent: true,
+      }),
+    ).toMatchInlineSnapshot(`
+            "SELECT
+              [Currency].[Currency].Members ON ROWS,
+              NON EMPTY {
+                [Measures].[contributors.COUNT]
+              } ON COLUMNS
+              FROM [EquityDerivativesCube]"
+          `);
+  });
+
+  it("returns the given MDX if a measure is already expressed", () => {
+    const mdx: MdxSelect = parse(`
+      SELECT
+        [Currency].[Currency].Members ON ROWS,
+        [Measures].[pnl.SUM] ON COLUMNS
+        FROM [EquityDerivativesCube]`);
+    expect(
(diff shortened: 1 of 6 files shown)


NZepeda (approved):
Nice, looks good!

src/4.3_to_5.0/_addDefaultMeasureIfNoneIsExplicitlyExpressed.test.ts
```diff
@@ -0,0 +1,44 @@
+import { MdxSelect, parse, stringify } from "@activeviam/mdx-5.0";
+import { _addDefaultMeasureIfNoneIsExplicitlyExpressed } from "./_addDefaultMeasureIfNoneIsExplicitlyExpressed";
+import { dataModelsForTests } from "@activeviam/data-model-5.0";
+
+const cube = dataModelsForTests.sandbox.catalogs[0].cubes[0];
+
+describe("_addDefaultMeasureIfNoneIsExplicitlyExpressed", () => {
```
Nouzbe, line 7: It would be nice to add a test ensuring that the function does not early-return by mistake when the query does not contain any measure but a hierarchy is sorted. For instance:

```
WITH
 Member [Measures].[[Currency]].[Currency]].[Currency]]_for_order] AS Rank(
  [Currency].[Currency].CurrentMember,
  [Currency].[Currency].Members
) 
SELECT
  NON EMPTY Order(
    Hierarchize(
      Descendants(
        {
          [Currency].[Currency].[ALL].[AllMember]
        },
        1,
        SELF_AND_BEFORE
      )
    ),
    [Measures].[[Currency]].[Currency]].[Currency]]_for_order],
    DESC
  ) ON ROWS
  FROM [EquityDerivativesCube]
```

I expect `[Measures].[contributors.COUNT]` to be added in this query.

Nouzbe (commented):
I just tried to reproduce in AUI4 and I actually got no measures shown either:

image

Can you reproduce?

NZepeda:
> I just tried to reproduce in AUI4 and I actually got no measures shown either:
> 
> image
> 
> Can you reproduce?

@Nouzbe This happens only for hierarchies on columns, if there is a hierarchy on columns, but no measure expressed, the default measure is shown on AUI-4

Screenshot 2024-10-01 at 2 27 43 PM

src/4.3_to_5.0/_addDefaultMeasureIfNoneIsExplicitlyExpressed.ts
```diff
@@ -0,0 +1,51 @@
...
+  { cube }: { cube: Cube },
+): MdxSelect | MdxDrillthrough;
+
+/**
+ * When there is any hierarchy expressed on axes, but no measure explicitly expressed, the default measure is:
+ * - displayed in Atoti UI 4.3
+ * - not displayed in Atoti 5.0
+ * This function explicitly adds it to the MDX so that it is isofunctional for the user.
+ */
+export function _addDefaultMeasureIfNoneIsExplicitlyExpressed(
```
NZepeda, line 29: Given that the default measure is only displayed when there is a hierarchy on columns with no explicit measures, it might be good to modify this function so that the measure is only explicitly added to the query in that scenario. Right now, it adds the measure if there is a hierarchy expressed in either the `rows` or `columns` axis, which is technically not isofunctional to what AUI-4 does.

NZepeda: See the behavior of AUI-4 below: 

https://github.com/user-attachments/assets/16253bcc-faff-4adc-b478-f1c05e15f73d

NZepeda:
I've added some commits to this PR to align the migration script's functionality with that of AUI-4. On AUI-4, the default measure was **only** added when no measures were expressed and the query contained at least hierarchy on the columns axis. The default measure was **not** added when the columns axis was empty. See this [video](https://github.com/activeviam/atoti-ui-migration/pull/130#discussion_r1787833121) for details. @Nouzbe @antoinetissier Let me know if you agree/disagree with the changes. We can drop the new commits if you disagree with the changes.

antoinetissier (commented):
Thanks @NZepeda 👍

PR #131, opened 2024-10-01
UI-9817 - Update dependency versions

NZepeda:
This PR bumps the version of the `atoti-ui-sdk` dependencies.


jest.config.js
```diff
@@ -2,7 +2,13 @@
 const isCI = require("is-ci");
 
 const extensions = ["js", "jsx", "ts", "tsx", "mjs", "json"];
-const esmPackages = ["@activeviam/*", "lodash-es", "monaco-editor"];
+const esmPackages = [
+  "@activeviam/*",
+  "lodash-es",
+  "monaco-editor",
+  "react-dnd-html5-backend",
+  "array-move",
```
NZepeda, line 10: Without the addition of these two, the following error was being thrown in the tests (notice the path being `node_modules/{packageName}`:

```
Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/nestorzepeda/dev/activeviam/atoti-ui-migration/node_modules/react-dnd-html5-backend/dist/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import { HTML5BackendImpl } from './HTML5BackendImpl.js';
                                                                                      ^^^^^^
```

